PR #43, opened 2024-09-04
Fix for Centos and Rocky VMs

tanaypf9:
Also contains fix for https://issues.redhat.com/browse/RHEL-54377
Parallelized incremental copy


jeremymv2 (approved):
Thank you for the improvements @tanaypf9!

PR #48, opened 2024-10-01
Add UI to Repo

tanaypf9:


vjailbreak-ui/.gitignore
```diff
@@ -0,0 +1,27 @@
+# Logs
```
jeremymv2, line 1: In the interest of saving on typing and eliminating redundancy from the name, I would prefer if we could have this simply under a directory named `ui`.

jeremymv2 (changes requested):

vjailbreak-ui/.gitignore
tanaypf9: Resolved

jeremymv2 (approved):
Nice addition!

PR #51, opened 2024-10-01
UI: Fetch the k8s API token from the pod

tanaypf9:
Dynamically fetch the k8s API token from the pod

### DIFF
--- a/ui/startup.sh
+++ b/ui/startup.sh
@@ -1,0 +1,8 @@
+#!/bin/sh
+
+# Use envsubst to substitute environment variables in the nginx configuration
+VITE_API_TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token) envsubst < /usr/share/nginx/html/index.html > /usr/share/nginx/html/index.html.temp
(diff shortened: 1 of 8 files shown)


ui/startup.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+# Use envsubst to substitute environment variables in the nginx configuration
+VITE_API_TOKEN=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token) envsubst < /usr/share/nginx/html/index.html > /usr/share/nginx/html/index.html.temp
+cp -f /usr/share/nginx/html/index.html.temp /usr/share/nginx/html/index.html
```
jeremymv2, line 5: I am partial to a `mv` here instead of a copy. This would be marginally cleaner since it eliminates the unused file on the system.

jeremymv2 (changes requested):

ui/startup.sh
tanaypf9: Changed cp to mv

jeremymv2 (approved):

PR #56, opened 2024-10-16
Add granular mapping options for volume type, networks and ports

### LINKED ISSUE #44: [Feature Request] Add granular volume type mapping
Current implementation requires mapping a data store to openstack volume type, which is great for most use-case. But for few advanced migrations, I would love to map individual virtual machine disks to different volume types.

It's more work during the migration, but provides us a greater flexibility to tier our workloads.

tanaypf9:
Fixes #44


jeremymv2 (approved):
Great addition @tanaypf9.

PR #58, opened 2024-10-28
Support for VMs with capital names

tanaypf9:
Add missing errors to log output

### DIFF
--- a/k8s/migration/internal/controller/migrationplan_controller.go
+++ b/k8s/migration/internal/controller/migrationplan_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"regexp"
 	"slices"
 	"strconv"
 	"strings"
@@ -31,6 +32,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/validation"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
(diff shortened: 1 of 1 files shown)


k8s/migration/internal/controller/migrationplan_controller.go
```diff
@@ -240,7 +240,7 @@ func (r *MigrationPlanReconciler) UpdateMigrationPlanStatus(ctx context.Context,
 func (r *MigrationPlanReconciler) CreateMigration(ctx context.Context,
 	migrationplan *vjailbreakv1alpha1.MigrationPlan,
 	vm string) (*vjailbreakv1alpha1.Migration, error) {
-	vmname := strings.ReplaceAll(strings.ReplaceAll(vm, " ", "-"), "_", "-")
+	vmname := strings.ToLower(strings.ReplaceAll(strings.ReplaceAll(vm, " ", "-"), "_", "-"))
```
jeremymv2, line 243: @tanaypf9 while I think this is a good start, somewhat surprisingly, it will not be sufficient for all real world scenarios. Looking at some real client data, I see some VMs begin with '_', while another VM has '(' and ')' in the name. The K8s object names have to be RFC1123 compliant and so those VM migrations would fail. 

Perhaps we can utilize the [upstream validation library](https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go) for inspiration to ensure the names we create are compliant?

jeremymv2 (changes requested):
@Tanay, while I think this is a great start, I see from some client data that some VM

k8s/migration/internal/controller/migrationplan_controller.go
tanaypf9: Created a new function to address this. Was able to rename a VM from `_M!g_(test]_©B*t_baK-c!0N€-9` to `mg-test-bt-bak-c0n-9`. I also use the same upstream validation library to check my generated name and return the same error that k8s would return in case some edge case is identified in the future.

jeremymv2 (approved):

PR #70, opened 2024-11-16
Minimal Tool to export VM info from VMware

tanaypf9:


jeremymv2 (approved):
Great addition @tanaypf9!

Issue #132, opened 2024-12-27
Need Additional VM details as `Running Status`, `Current IP`, `OS`, `Network Info`, `Cluster Name` in the migrations API


patil-pratik-87:
I need these details to implement #106  

For ref --> ![Image](https://github.com/user-attachments/assets/80c7e148-4abb-48e0-98d0-7328f80d634e)

OmkarDeshpande7:
@patil-pratik-87 Can you share entire screen's screenshot instead? I want to see the column names as well

patil-pratik-87:
It got deleted while we were toying with Figma :) .
The columns are =>  `VM Name`, `Running Status`, `Current IP`,` Network Info`, `OS`, `vCenter Cluster Name`

OmkarDeshpande7:
@patil-pratik-87 I am not sure last 3 columns are needed for display to the user. There are technical issues fetching with those.
- `Network Info` - which network? finding primary network from all the networks is tedious
- `OS` - we can determine OS_TYPE at this point without actually going inside each VM (which maybe thousands of them)
- `vCenter Cluster Name` - Not presented in get VM call by vsphere

PR #133, opened 2024-12-30
add IpAddress and VM state info to migrationtemplate status

OmkarDeshpande7:
# This PR includes
- Adding IpAddress and VM state info to migrationtemplate status
- Adding helper make targets
- Resolving need to scale-down and scale-up controller manager deployment every time a new image is created. Done by setting `leader-election` to `false` which was anyway ineffective since we have only one replica

#132


jeremymv2 (approved):
Looks good!

PR #142, opened 2025-01-07
141 UI handle vm refresh with additional put call to migration template and addadjust the timeout needed for it to reflect latest detail

patil-pratik-87:
https://github.com/user-attachments/assets/91402e24-2e85-44a2-9cd4-553829817fcb

Changes ->

1. Make the dummy PATCH call for the migration template to trigger reconciliation in order to get latest VM details on manual refresh.
2. Added a 20-second forced timeout for every manual refresh, as it takes time for the latest VM details to reflect.
3. Disabled migration with a tooltip informing user to refresh, if the VM does not have an IP.

### DIFF
--- a/ui/src/features/migration/VmsSelectionStep.tsx
+++ b/ui/src/features/migration/VmsSelectionStep.tsx
@@ -81,6 +81,7 @@ const columns: GridColDef[] = [
 const paginationModel = { page: 0, pageSize: 5 };
 
 const DISABLED_TOOLTIP_MESSAGE = "Turn on the VM to enable migration.";
+const NO_IP_TOOLTIP_MESSAGE = "VM is running but does not have an IP, try refreshing.";
 
 interface VmsSelectionStepProps {
   vms: VmData[];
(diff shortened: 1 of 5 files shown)


jeremymv2 (approved):

ui/src/features/migration/VmsSelectionStep.tsx
```diff
@@ -81,6 +81,7 @@ const columns: GridColDef[] = [
 const paginationModel = { page: 0, pageSize: 5 };
 
 const DISABLED_TOOLTIP_MESSAGE = "Turn on the VM to enable migration.";
+const NO_IP_TOOLTIP_MESSAGE = "VM is running but does not have an IP, try refreshing.";
```
anmolsachan, line 84: VM is not assigned an IP yet. Please refresh again.

anmolsachan (changes requested):

ui/src/features/migration/VmsSelectionStep.tsx
anmolsachan, line 84:
```suggestion
const NO_IP_TOOLTIP_MESSAGE = "VM is not assigned an IP yet. Please refresh again.";
```

anmolsachan (changes requested):

ui/src/features/migration/VmsSelectionStep.tsx
anmolsachan: Looks good otherwise.

anmolsachan (approved):

PR #159, opened 2025-01-31
[packer optimisation for scale] Make packer to dynamically spawn as master node or worker node ( release )

spai-p9:
This PR adds dynamic k3s cluster provisioning using cloud-init for OpenStack migration. During the first boot, VMs automatically configure themselves as either master or worker nodes. The master sets up the k3s cluster, applies kube-prometheus manifests, and shares a token via an HTTP server, while workers use this token to join the cluster. This approach simplifies scaling, improves deployment flexibility, and keeps the setup clean by separating build-time (Packer) and runtime (cloud-init) tasks.


.github/workflows/packer.yml
```diff
@@ -163,7 +163,7 @@ jobs:
     - name: Download base image
       if: env.release_found == 'true'
       run: |
-          oras pull quay.io/platform9/vjailbreak:base
```
srinivaschavan, line 166: Lets make sure we publish a new image in our repo and use that.

spai-p9: Yeah still testing out the changes hence used private image, once done will push out a image as well.

.github/workflows/packer.yml
```diff
@@ -163,7 +163,7 @@ jobs:
...
+          oras pull docker.io/spaip9/vjailb:vjail
```
OmkarDeshpande7, line 166:
```suggestion
          oras pull docker.io/spaip9/vjailb:vjail
```
blocking merge so this is not checked in accidentally

image_builder/cloudinit/user-data
```diff
@@ -1,4 +1,65 @@
...
+chpasswd: 
+  expire: False
+ssh_pwauth: True
+
+package_update: true
+package_upgrade: all
+
+write_files:
+  # File to store k3s token (for worker retrieval)
+  - path: /etc/k3s-token
```
OmkarDeshpande7, line 12: as discussed the path should be `/etc/pf9/k3s/token`

image_builder/cloudinit/user-data
```diff
@@ -1,4 +1,65 @@
...
+    # Logging function
+    log() {
+      echo "[$(date '+%Y-%m-%d %H:%M:%S')] $1" | sudo tee -a /var/log/vjailbreak.log
+    }
+
+    # Start logging
+    log "Starting vjailbreak setup for master node..."
+
+    log "k3s not found. Installing k3s..."
+    curl -sfL https://get.k3s.io | sh -
```
OmkarDeshpande7, line 27: We are not sending user-data for primary node, so k3s server should be installed by default

OmkarDeshpande7 (changes requested):
feedback for user-data

image_builder/cloudinit/user-data
spai-p9: since we have moved out of this cloud-init approach not relevant anymore.

image_builder/cloudinit/user-data
spai-p9: not relevant since, new approach using cron job.

.github/workflows/packer.yml
spai-p9: removed this for now.

.github/workflows/packer.yml
```diff
@@ -163,7 +163,7 @@ jobs:
     - name: Download base image
       if: env.release_found == 'true'
       run: |
-          oras pull quay.io/platform9/vjailbreak:base
+          oras pull quay.io/spai-p9/vjailbreak:base
```
OmkarDeshpande7, line 166: please use platform9 image here

OmkarDeshpande7 (commented):
one comment, looks goos otherwise

.github/workflows/packer.yml
sharma-tapas: version this tag out to have v0.1.4-base

sharma-tapas (approved):

.github/workflows/packer.yml
spai-p9: Done

PR #160, opened 2025-02-01
Minor UI & readme updates

damian-pf9:
Just some minor UI updates such as VMWare to VMware, Openstack to OpenStack, etc.


OmkarDeshpande7:
@damian-pf9 Some checks are failing. Can you look into those?

PR #161, opened 2025-02-03
Adding workers nodes to the vjailbreak

OmkarDeshpande7:
This PR includes following major changes

- Addition of `VjailbreakNode` API for tracking worker nodes
- Controller for `VjailbreakNode`
- Upgrade controller-runtime version
- Refactoring v2v-helper


k8s/migration/pkg/constants/constants.go
```diff
@@ -0,0 +1,37 @@
...
+password: %s
+chpasswd: { expire: False }
+write_files:
+- path: %s
+  content: |
+    export IS_MASTER=%s
+    export MASTER_IP=%s
+    export K3S_TOKEN=%s
+runcmd:
+  - echo "Created k3s env variables!" > /home/ubuntu/cloud-init.log
```
spai-p9, line 35: I tried with this and the env variables didn't change,

```suggestion
	CloudInitScript = `#cloud-config
password: winterwonderland
chpasswd: { expire: False }
ssh_pwauth: True
manage_etc_hosts: true

runcmd:
  - sed -i '/^export IS_MASTER/d' /etc/pf9/k3s.env
  - sed -i '/^export MASTER_IP/d' /etc/pf9/k3s.env
  - sed -i '/^export K3S_TOKEN/d' /etc/pf9/k3s.env
  - echo "export IS_MASTER=false" >> /etc/pf9/k3s.env
  - echo "export MASTER_IP=%s" >> /etc/pf9/k3s.env
  - echo "export K3S_TOKEN=%s" >> /etc/pf9/k3s.env
runcmd:
  - echo "Created k3s env variables!" > /home/ubuntu/cloud-init.log
```

k8s/migration/pkg/constants/constants.go
```diff
@@ -0,0 +1,37 @@
...
+
+var (
+	// Cloud-Init Script (User Data)
+	CloudInitScript = `#cloud-config
+password: %s
+chpasswd: { expire: False }
+write_files:
+- path: %s
+  content: |
+    export IS_MASTER=%s
```
spai-p9, line 31: Also since this is for worker node let IS_MASTER be false by default no need to pass?

k8s/migration/pkg/constants/constants.go
```diff
@@ -0,0 +1,37 @@
...
+const (
+	VjailbreakNodeControllerName = "vjailbreaknode-controller"
+	K8sMasterNodeAnnotation      = "node-role.kubernetes.io/master"
+	NodeRoleMaster               = "master"
+	InternalIPAnnotation         = "k3s.io/internal-ip"
+
+	VjailbreakNodePhaseVMCreating  = vjailbreakv1alpha1.VjailbreakNodePhase("CreatingVM")
+	VjailbreakNodePhaseVMCreated   = vjailbreakv1alpha1.VjailbreakNodePhase("VMCreated")
+	VjailbreakNodePhaseNodeCreated = vjailbreakv1alpha1.VjailbreakNodePhase("NodeCreated")
+
```
spai-p9, line 14: Can we have one more phase where we are not sure about the vm state?

k8s/migration/pkg/constants/constants.go
OmkarDeshpande7: default value is passed in the format string

k8s/migration/pkg/constants/constants.go
OmkarDeshpande7: Sure, we can have that

k8s/migration/pkg/constants/constants.go
OmkarDeshpande7: Seems to be working. with the new image
```
export IS_MASTER=false
export MASTER_IP=10.9.3.95
export K3S_TOKEN=XYZ
```

sharma-tapas (approved):

PR #227, opened 2025-02-26
Support OS installed on LV

### LINKED ISSUE #146: boot disk for VM using LVs as boot volumes are not detected
If user has a LV spanning multiple physical devices and used as boot volume we cannot detect bootable disk

Disks of such VM are attached to the vjailbreak container
Screenshot 2025-01-16 at 2 25 31 PM

using disk 1 (vdb)
Screenshot 2025-01-16 at 2 26 15 PM

using both disks at the same time
Screenshot 2025-01-16 at 2 27 57 PM

OmkarDeshpande7:
Supports detecting OS installations, even if OS is installed on LVM. Find bootable disk, and create correct target VM on openstack.

Fixes #146


v2v-helper/migrate/migrate.go
```diff
@@ -376,102 +398,119 @@ func (migobj *Migrate) LiveReplicateDisks(ctx context.Context, vminfo vm.VMInfo)
...
+		}
+
+		openstackops := migobj.Openstackclients
+		err = openstackops.SetVolumeBootable(vminfo.VMDisks[bootVolumeIndex].OpenstackVol)
+		if err != nil {
+			return fmt.Errorf("failed to set volume as bootable: %s", err)
 		}
 	}
+
+	time.Sleep(10 * time.Minute)
```
OmkarDeshpande7, line 497: leftover debug?

spai-p9 (approved):

Issue #234, opened 2025-02-28
panic due to wrong OpenStack endpoint certificate validation


bhavin192:
### What steps did you take and what happened?

My OpenStack setup doesn't use https auth endpoint.

1. Uploaded the OpenStack RC file from my setup (it didn't have `OS_INSECURE` in it)
2. The file got validated without any errors.
3. But when migration controller tried to run `GetOpenstackInfo`, it resulted in a nil pointer dereference panic.

Logs:

```
2025-02-26T13:16:54Z    ERROR   Error getting certificate for 'http://192.168.4.202:5000'       {"controller": "migrationtemplate", "controllerGroup": "vjailbreak.k8s.pf9.io", "controllerKind": "MigrationTemplate", "MigrationTemplate": {"name":"5c7ef3e9-5890-46a4-941c-5a622156bd83","namespace":"migration-system"}, "namespace": "migration-system", "name": "5c7ef3e9-5890-46a4-941c-5a622156bd83", "reconcileID": "5eecbeb3-0163-4818-8b0e-647bbf07e678"}
github.com/platform9/vjailbreak/k8s/migration/internal/controller.validateOpenstackCreds
        /workspace/internal/controller/openstackcreds_controller.go:155
github.com/platform9/vjailbreak/k8s/migration/internal/controller.GetOpenstackInfo
        /workspace/internal/controller/openstackcreds_controller.go:314
github.com/platform9/vjailbreak/k8s/migration/internal/controller.(*MigrationTemplateReconciler).Reconcile
        /workspace/internal/controller/migrationtemplate_controller.go:91
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Reconcile
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.1/pkg/internal/controller/controller.go:118
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.1/pkg/internal/controller/controller.go:328
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem
        /go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.1/pkg/internal/controller/controller.go:288
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2
[...]

PR #235, opened 2025-02-28
Fix panic caused by OpenstackCreds CA cert check

bhavin192:
Currently, the check passes even when there is a failure to get certificate from the endpoint. This is due to a typo in error check, we end up checking for wrong err variable.

Because of it ends up with a nil pointer dereference panic in GetOpenstackInfo when it is called by migration controller to get details of volume etc.

Test plan:

1. Built a new controller image.
2. Uploaded a OpenStack RC file with HTTP auth URL and without OS_INSECURE.
3. The validation failed as expected.

Screenshot 2025-02-28 at 16-44-33 vJailbreak

Fixes https://github.com/platform9/vjailbreak/issues/234

### DIFF
--- a/k8s/migration/internal/controller/openstackcreds_controller.go
+++ b/k8s/migration/internal/controller/openstackcreds_controller.go
@@ -163,8 +163,8 @@ func validateOpenstackCreds(ctxlog logr.Logger, openstackcreds *vjailbreakv1alph
 		// Get the certificate for the Openstack endpoint
 		caCert, certerr := getCert(openstackCredential.AuthURL)
 		if certerr != nil {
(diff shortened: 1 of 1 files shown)


OmkarDeshpande7 (approved):
Looks Good

bhavin192:
Thank you for taking a look @OmkarDeshpande7, I just rebased this PR against latest changes. Can you run the workflow which needs approval?

OmkarDeshpande7 (approved):

sharma-tapas (approved):

sharma-tapas (approved):

Issue #285, opened 2025-03-17
Implement a seperate credential creation workflow for vmware and openstack creds, and remove the creation capability from the MigrationForm


patil-pratik-87:
### Is your feature request related to a problem? Please describe?

.

### Detailed Description

.

### Anything else you would like to add?

.

PR #298, opened 2025-03-20
UI: Implement a new credential addition workflow, and remove the creds addition from MigrationForm and Scaleup drawer

patil-pratik-87:
#285


spai-p9 (approved):
lgtm

PR #558, opened 2025-06-11
fix openstack volume sizes being smaller than original disk size

### LINKED ISSUE #557: Volumes created on Openstack are of less size than original disk size on vcenter

Volumes created on Openstack are of less size than that of original vcenter disk size. This makes copy phase to fail with nbdcopy error.
```
nbdcopy: error: destination size (x) is smaller than source size (y)
```

For example, if actual volume size is 100.25 GB, then the volume created on openstack is of 100GB only

nbdcopy to execute successfully.

### Environment

v0.1.12

### vCenter version

All

_No response_

OmkarDeshpande7:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #557 

This PR fixes issue by adding 1 GB extra space while creating volumes. This is to accommodate any leftover bytes during size calculation. 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR enhances volume creation by updating size calculation with math.Ceil and adding 1GB to address leftover bytes. It also refactors VM operations code by removing obsolete comments and adding logic to collect detailed virtual disk data, improving reliability in disk sizing and code maintainability.

</div>

### DIFF
--- a/v2v-helper/pkg/utils/openstackopsutils.go
+++ b/v2v-helper/pkg/utils/openstackopsutils.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


spai-p9 (approved):
lgtm

PR #611, opened 2025-06-20
do not fail migrations for snapshot delettions

### LINKED ISSUE #218: snapshot discovery resolved to two different snapshot

Steps:

- Start a migration
- delete the kubernetes node having pod for that migration. (basically do an abrupt migration somehow)
- re-do the migration for same VM.

You will see an error that says snapshot reference "migration-snap" resolved to two snapshots, and migration will fail.

Ideally the second migration should have continued since this is a new snapshot at a different time.

### Environment

v0.1.4

### vCenter version

any

_No response_

### LINKED ISSUE #610: Backend: Handle snapshot creation and deletion more gracefully

Now before we start migration we do not check if the snapshot exists or not, and after the copy is done and if the snapshot deletion fails We fail the snapshot completely, which is not at all a good behaviour. because this failure of  deletion of snapshot is not at all harmful. 

All of these to be handled correctly

### Environment

v0.1.13

### vCenter version

NA

_No response_

OmkarDeshpande7:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
fixes #610 
fixes #218


v2v-helper/migrate/migrate.go
```diff
@@ -229,8 +229,15 @@ func (migobj *Migrate) LiveReplicateDisks(ctx context.Context, vminfo vm.VMInfo)
 		}
 	}
 
+	// clean up snapshots
+	utils.PrintLog("Cleaning up snapshots before copy")
+	err := vmops.CleanUpSnapshots(false)
```
spai-p9, line 234: @OmkarDeshpande7 So we are gonna fail here if we can't clean up snapshot?

v2v-helper/migrate/migrate.go
```diff
@@ -229,8 +229,15 @@ func (migobj *Migrate) LiveReplicateDisks(ctx context.Context, vminfo vm.VMInfo)
...
+	if err != nil {
+		return vminfo, fmt.Errorf("failed to clean up snapshots: %s", err)
```
spai-p9, line 236: Can we make the error like, this:

```suggestion
		return vminfo, fmt.Errorf("failed to clean up snapshots: %s, Please delete manually before starting again,", err)
```

something like this:

v2v-helper/vm/vmops.go
```diff
@@ -476,3 +506,75 @@ func (vmops *VMOps) VMPowerOn() error {
...
+	var lastError error
+	snapshotsDeleted := 0
+
+	// Helper function to recursively process snapshot tree
+	var processSnapshotTree func(trees []types.VirtualMachineSnapshotTree) int
+	processSnapshotTree = func(trees []types.VirtualMachineSnapshotTree) int {
+		deleted := 0
+		for _, snapshot := range trees {
+			// First process any child snapshots
+			if len(snapshot.ChildSnapshotList) > 0 {
```
spai-p9, line 535: We can check if snapshot.Name == constants.MigrationSnapshotName before going into the recursive loop. This way, we only go through snapshot chains that start with migration-snap and skip others that don’t matter.

It helps avoid unnecessary traversal and keeps the logic focused on just the snapshots we want to delete.

spai-p9 (approved):
lgtm

Issue #1020, opened 2025-10-16
Admin Cutover incremental sync


spai-p9:
### Is your feature request related to a problem? Please describe?

Currently, when a user selects Admin Cutover and initiates a migration in vJailbreak, the system copies all the volumes and then waits for the user to trigger the final cutover. During this waiting period, no further data synchronization occurs. If the cutover is delayed for an extended time, a large number of changed blocks accumulate, resulting in a heavy one-time sync operation during the cutover phase. This can significantly increase cutover duration and impact overall migration efficiency.

### Detailed Description

Introduce a mechanism that periodically syncs incremental changes between the source and target environments after the initial full data copy is completed. This background sync process should run at configurable intervals, continuously replicating any changed blocks while the system awaits the user’s cutover action.

When the final cutover is triggered, only the minimal delta (the last set of changes) would need to be copied, reducing downtime and improving migration speed and consistency.

meghansh-pf9:
Hi, CBT configuration for the sync interval will be managed manually via ConfigMap. We will also provide an option to configure this from the ui while figuring the migration
Moving forward, we are developing an automated sync algorithm to optimize performance. The long-term design includes a user-facing configuration to select the preferred mode: manual configuration or algorithmic scheduling.

PR #1056, opened 2025-10-30
Admin Cutover Periodic sync

meghansh-pf9:

**Scheduled Synchronization:** Syncs now run based on a configurable time interval in the vjailbreak-settings configmap, moving away from continuous looping.

**Conditional Preemption Logic:**

- If the label is set to "yes" while waiting for the next schedule, the wait is terminated, and an expedited cutover is initiated immediately.

- If the label is set to "yes" while a CBT is running, the current CBT must complete to preserve data integrity, and then the expedited cutover is initiated.

CBT Integrity Assurance: Logic is implemented to prevent CBT data corruption by ensuring any running sync completes before initiating the final cutover, as detailed in the design document.

fixes: #1020 

__Testing__

### Configmap Configuration:
image

### Waiting for admin cutover state:
image

### Interval Exhausted:
image

### Periodic Sync Running:
image 
 <div id='description'>
<h3>Summary by Bito</h3>

<ul>

<li>This pull request introduces a configurable periodic sync interval, moving away from continuous looping, which may impact synchronization reliability.</li>

<li>It adds conditional preemption logic to ensure data integrity during cutover operations, potentially affecting system behavior under certain conditions.</li>

<li>The changes enhance resource management, allowing for more flexible synchronization processes.</li>

<li>Overall, the modifications focus on improving synchronization configurations and ensuring safe cutover operations, introducing risks related to synchronization reliability.</li>

</ul>

</div>


v2v-helper/migrate/migrate.go
```diff
@@ -240,16 +240,195 @@ func (migobj *Migrate) WaitforCutover() error {
...
-			break
+		migobj.logMessage("DEBUG : Waiting for Admin Cutover conditions to be met")
+		select {
+		case label := <-migobj.PodLabelWatcher:
+			if label == "yes" {
+				// wait for sync to finish
+				for syncRunning {
+					continue
+				}
+				break outLoop
```
spai-p9: Instead of outloop, can't we just return?

v2v-helper/migrate/migrate.go
```diff
@@ -240,16 +240,195 @@ func (migobj *Migrate) WaitforCutover() error {
...
+				case <-ctx.Done():
+					err = vmops.CleanUpSnapshots(false)
+					changedBlockCopySuccess = false
+					if err != nil {
+						syncChan <- errors.Wrap(err, "failed to cleanup snapshot of source VM")
+					}
+				default:
+					syncChan <- errors.Wrap(err, "failed to copy changed blocks")
+				}
+				*syncRunning = false
```
spai-p9: This is probably not needed because if we add the defer, irrespective of success or failure we would need to make this false, hence no need to set it everywhere

v2v-helper/migrate/migrate.go
```diff
@@ -240,16 +240,195 @@ func (migobj *Migrate) WaitforCutover() error {
...
+	if *intervalExhausted && !*syncRunning {
+		*intervalExhausted = false
+		migobj.logMessage("Previous interval completed and no sync in progress - rescheduling next sync")
+		go migobj.SetInterval(intervalExhausted)
+		return true
+	}
+	return false
+}
+
+func (migobj *Migrate) WaitforAdminCutover(vminfo vm.VMInfo) error {
```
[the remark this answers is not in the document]
meghansh-pf9: removed the new context creation in simplified logic

k8s/migration/internal/controller/migrationplan_controller.go
```diff
@@ -1441,10 +1441,6 @@ func (r *MigrationPlanReconciler) migrateRDMdisks(ctx context.Context, migration
 				if err != nil {
 					return fmt.Errorf("failed to validate RDMDisk CR: %w", err)
 				}
-				// collect RDM disks that need to be updated (to be imported to Cinder)
-				if !rdmDiskCR.Spec.ImportToCinder {
-					rdmDiskCRToBeUpdated = append(rdmDiskCRToBeUpdated, *rdmDiskCR)
-				}
```
spai-p9: Looks like for some reason this is getting removed in your branch, checked in main this piece of code still exists, please check and add this back.

v2v-helper/migrate/migrate.go
```diff
@@ -240,17 +240,164 @@ func (migobj *Migrate) WaitforCutover() error {
...
+				err = migobj.SyncCBT(ctx, vminfo)
+				if err != nil {
+					return err
+				}
+			}
 		}
 	}
-	migobj.logMessage("Cutover conditions met")
-	return nil
 }
```
spai-p9:
```suggestion
func (migobj *Migrate) WaitforAdminCutover(ctx context.Context, vminfo vm.VMInfo) error {
	migobj.logMessage("Waiting for Admin Cutover conditions to be met")

	syncInterval := migobj.getSyncDuration()

	for {
		select {
		case <-ctx.Done():
			return ctx.Err()

		case label := <-migobj.PodLabelWatcher:
			if label == "yes" {
				migobj.logMessage("Admin cutover triggered")
				return nil
			}

		default:
			// Perform sync
			migobj.logMessage(fmt.Sprintf("Starting periodic sync (interval: %s)", syncInterval))
			start := time.Now()
			if err := migobj.SyncCBT(ctx, vminfo); err != nil {
				return err
			}
			elapsed := time.Since(start)

			// If sync took longer than interval → run next sync immediately
			if elapsed >= syncInterval {
				migobj.logMessage("Sync took longer than interval → immediately starting next cycle")
				continue
			}

			// Otherwise wait remaining time
			waitTime := syncInterval - elapsed
			migobj.logMessage(fmt.Sprintf("Sync finished early → waiting %s before next sync", waitTime))

			select {
			case <-ctx.Done():
				return ctx.Err()
			case <-migobj.PodLabelWatcher:
				return nil // admin triggered cutover during wait
			case <-time.After(waitTime):
				// wait completed → loop and sync again
			}
		}
	}
}
```
@meghansh-pf9 can i suggest this change? We can simplify this logic by removing SetInterval() and DecideReschedule(), and instead use a time.After wait directly after each sync. This avoids starting a background goroutine to flip a shared boolean flag, which can lead to race conditions and makes the flow harder to follow. With the time.After approach, the loop becomes self-contained, readable, and ensures we never overlap sync runs while still immediately starting the next sync if the previous one exceeded the interval. Overall, this makes the periodic sync logic cleaner and easier to maintain. if that makes sense to you please feel free to use this or make changes to however you feel right.

spai-p9 (changes requested):

v2v-helper/migrate/migrate.go
meghansh-pf9: Keeping the getSync Interval outside the loop will make it less configurable, if user wants to change the interval while the periodic sync is running it won't be possible. Instead we can keep it in the loop in case of updations.

OmkarDeshpande7 (approved):
LGTM

PR #1115, opened 2025-11-14
Added Pre-migration check to see if port is available before migration

### LINKED ISSUE #1089: Add Pre-migration Validation for PCD Port/IP Conflicts

Start Migration

Wait for copying the VM's disk

The migration failed with an error: "Port is already in use."

The Problem: You didn't realize that another VM was already running on the destination (PCD) with that same IP, so the new VM wasn't allowed to start.

Implement a precheck to validate if a VM's IP address is already in use by a port on the PCD. This would provide a "fail-fast" mechanism, preventing migration failures after the time-consuming vmdk data copy is complete.

### Environment

v0.3.5

### vCenter version

NA

_No response_

sarika-pf9:

fixes #1089

## Testing done

image
image
 
 <div id='description'>
<h3>Summary by Bito</h3>

<ul>

<li>This pull request introduces pre-migration checks to verify if a port is available before proceeding with the migration, aiming to prevent potential conflicts during the migration process.</li>

<li>Minor refactoring is included in the controller and utility files to improve code readability and structure.</li>

<li>The changes aim to increase the reliability of the system during migration.</li>

<li>Overall, the pull request introduces pre-migration checks and includes minor refactoring in the controller and utility files.</li>

</ul>

</div>


spai-p9 (approved):
Just a thought here: if there is a deviceID that means on pcd there is a VM, now when we error out here, by default if we see a error we power on VM on the source, we should think about keeping the VM on the state it was before migration, before or after. we can take that up in a different PR. but it is a important change I believe.

PR #1314, opened 2025-12-27
Added advanced option in UI to select if the user wants to persist network interfaces

### LINKED ISSUE #1253: Ip assigned to port created on openstack for a VM should be reflected inside the vm statically for all OS

Currently, after VM migration, the IP address allocated to the port in OpenStack is correctly reflected inside Ubuntu guests that use netplan (the configuration is automatically applied inside the VM post-migration).
For any other OS only the port is created and inside the vm it is not reflected.

### Detailed Description

We need to extend this same behavior to all other supported guest operating systems (e.g., CentOS/RHEL, Debian without netplan, Windows, SUSE, etc.), so that the OpenStack-assigned port IP is automatically becomes the active IP configuration inside the guest, regardless of the OS or its native network configuration method.

_No response_

sarika-pf9:
fixes #1253 

## Testing done
image
image
image

### DIFF
--- a/v2v-helper/pkg/utils/utils.go
+++ b/v2v-helper/pkg/utils/utils.go
@@ -177,5 +177,6 @@ func GetNetworkPersistance(ctx context.Context, client client.Client) bool {
 	if err != nil {
 		return false
 	}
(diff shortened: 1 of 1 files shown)


meghansh-pf9 (approved):
LGTM

PR #1349, opened 2026-01-08
Handled trailing slash in vCenter URL

### LINKED ISSUE #1306: vcenter URL does not work if there is a / in the end

The vCenter URL does not work if it ends with a trailing “/”. The code should handle this by trimming any trailing “/” before using the URL

User should be able to provide url with "/" at the end

https://vcenter.phx.pnap.platform9.horse/

User should be able to provide url with "/" at the end

### Environment

v0.3.8-rc

### vCenter version

NA

_No response_

sarika-pf9:
This PR addresses an issue where providing a vCenter URL with a trailing slash (e.g., https://vcenter.example.com/) caused connection failures. The existing logic was blindly appending /sdk to the host string, resulting in malformed URL paths like //sdk, which triggered a 400 Bad Request from the vCenter API.

fixes #1306 

## Testing done

https://github.com/user-attachments/assets/9bfed29a-e357-41c1-b034-7e44cfa53c3b

### DIFF
--- a/v2v-helper/nbd/nbdops.go
+++ b/v2v-helper/nbd/nbdops.go
@@ -83,6 +83,8 @@ var fixedOptArgs = libnbd.BlockStatusOptargs{
 }
 
 func (nbdserver *NBDServer) StartNBDServer(vm *object.VirtualMachine, server, username, password, thumbprint, snapref, file string, progchan chan string) error {
+	server = strings.TrimRight(server, "/")
+
 	tmp_dir, err := os.MkdirTemp("", "nbdkit-")
 	if err != nil {
 		return fmt.Errorf("failed to create temp dir: %v", err)
@@ -175,7 +177,7 @@ func (nbdserver *NBDServer) CopyDisk(ctx context.Context, dest string, diskindex
 	utils.PrintLog(fmt.Sprintf("Executing %s\n", cmdString))
 	go func() {
 		scanner := bufio.NewScanner(progressRead)
-		
+
 		lastLoggedProgress := -1
 		const logInterval = 5
 		lastChannelProgress := 0


k8s/migration/pkg/utils/credutils.go
```diff
@@ -472,16 +472,18 @@ func ValidateVMwareCreds(ctx context.Context, k3sclient client.Client, vmwcreds
...
+	if !strings.HasPrefix(rawHost, "http") {
+		rawHost = "https://" + rawHost
 	}
-	u, err := url.Parse(host)
+
+	u, err := url.Parse(rawHost)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse URL: %w", err)
 	}
+	u.Path = "/sdk"
```
roopakparikh: dont use string concatenation. Use url.JoinPath

pkg/validation/vmware/validate.go
```diff
@@ -91,14 +91,23 @@ func Validate(ctx context.Context, k8sClient client.Client, vmwcreds *vjailbreak
 	datacenter := vmwareCredsinfo.Datacenter
 
 	// Normalize URL
-	if host[:4] != "http" {
-		host = "https://" + host
+	rawHost := strings.TrimSpace(host)
```
roopakparikh: This is copy pasted code from the code in the file above, why do we have similar looking functions, why not consolidate

pkg/vpwned/sdk/targets/vcenter/vcenter.go
```diff
@@ -74,10 +74,16 @@ func findBootDevice(devices []types.BaseVirtualMachineBootOptionsBootableDevice)
 func (v *Vcenter) connect(ctx context.Context, a api.TargetAccessInfo) (*govmomi.Client, *find.Finder, error) {
 	logrus.Info("Connecting to vCenter...")
 	// Parse vCenter URL
-	u, err := url.Parse(fmt.Sprintf("https://%s/sdk", a.HostnameOrIp))
+	rawHost := strings.TrimSpace(a.HostnameOrIp)
```
roopakparikh: Same again ? Why do we hav ethe duplication ?

k8s/migration/pkg/utils/credutils.go
sarika-pf9: made changes accordingly

PR #1422, opened 2026-01-28
fix: added buildBackendToVolumeTypeMap function and cleanup logic

### LINKED ISSUE #1420: Auto-discovery of NetApp storage array credentials is incorrect

Steps -

1. Add pcd creds
2. Check for discovered array creds

Image

It should be netapp/netapp but we are getting cinder_vol1_netapp

It should detect correct creds

### Environment

v0.3.8-rc

### vCenter version

NA

_No response_

### LINKED ISSUE #1421: Custom volume backend is not discovered properly with storage array credentials

Custom volume backend is not discovered properly with storage array credentials

1. Added custom volume backend in blueprint which is configured in cinder_override.conf

Couldn't discover vendor details for same

Vendor details should be discovered for custom volume backend

### Environment

v0.3.8-rc

### vCenter version

NA

_No response_

### LINKED ISSUE #1423: [Storage Accelerated Copy] Migration fails with - Failed to clone disk: The file already exists (39)

[Storage Accelerated Copy] Migration fails with - Failed to clone disk: The file already exists (39)

Whenever a migration fails for any reason after the disk copy is completed using Storage Accelerated Copy, the subsequent migration or retry fails with the error:
“Failed to clone disk: The file already exists (39)”

We should handle this with a new name or delete old one.

We should handle this error with a new name or delete old one.

### Environment

v0.3.8-rc

### vCenter version

NA

_No response_

sanya-pf9:
Added buildBackendToVolumeTypeMap function (before line 2074)
This function queries Cinder's volume types API and extracts the
volume_backend_name from each volume type's ExtraSpecs. This is the authoritative mapping that Cinder itself uses.

fixes #1420
fixes #1423 
fixes #1421

image

image

### DIFF
--- a/ui/src/components/layout/Sidenav/Sidenav.tsx
+++ b/ui/src/components/layout/Sidenav/Sidenav.tsx
@@ -399,6 +399,13 @@ export default function Sidenav({
   }
 
   const handleItemClick = (item: NavigationItem) => {
+    if ((window as any).__VDDK_UPLOAD_IN_PROGRESS__) {
+      const confirmed = window.confirm(
+        'File upload is in progress. If you leave this page, the upload may be interrupted. Do you want to continue?'
+      )
+      if (!confirmed) return
+    }
+
     if (onItemClick) {
       onItemClick(item)
     } else if (item.external) {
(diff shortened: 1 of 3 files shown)


spai-p9 (approved):
lgtm

Issue #1450, opened 2026-02-02
UI is not allowing to delete auto discovered storage mapping


geet-pf9:
### What steps did you take and what happened?

UI is not allowing to delete auto discovered storage mapping

https://github.com/user-attachments/assets/809742b3-8545-468b-9c7e-b4145ef8857e

### What did you expect to happen?

User should be able to delete the mapping and select based on user preference

### Environment

v0.4.0-rc

### vCenter version

NA

PR #1470, opened 2026-02-05
#1450 :: UI is not allowing to delete auto discovered storage mapping

AbhijeetThakur:
- Prevent re-adding auto-mapped array credentials after manual removal

fixes #
- https://github.com/platform9/vjailbreak/issues/1450

## Testing done
Dev testing is pending due to the unavailability of the setup.

image
tested this


spai-p9 (approved):
lgtm

PR #1511, opened 2026-02-17
Feature/option not to preserve ip and mac

### LINKED ISSUE #1487: Request option to NOT preserve the IP address and MAC address to allow migration to a different subnet

We have VMs that need to move to a different subnet as part of the migration for various reasons. We would like the option to tell vJailbreak not to preserve IP and MAC but instead just accept a generated MAC and new IP from the available allocation pool in a designated subnet in PCD.

### Detailed Description

When configuring a migration there should be an option to override the default behavior of IP retention. When selected vJailbreak would only need to prompt for the destination network (mapping source to destination wouldn't be required but shouldn't hurt either).

It would be appropriate to warn when selecting that option that it will result in the VM being assigned a new IP address and MAC address in PCD. It would also be helpful if vJailbreak verified that there is an IP available in the destination subnet allocation pool when configuring the migration (before submission).

sanya-pf9:
When configuring a migration there should be an option to override the default behavior of IP retention. When selected vJailbreak would only need to prompt for the destination network (mapping source to destination wouldn't be required but shouldn't hurt either).

fixes #1487 

image
image


spai-p9 (approved):
lgtm

PR #1549, opened 2026-02-23
fix: scope VM validation to non-terminal states and harden post-migration actions

### LINKED ISSUE #1556: Fix Reconciliation Failures and Incomplete Post-Migration Actions in Multi-VM Migration Plans
Image
When migrating multiple VMs in a single MigrationPlan with post-migration actions enabled (e.g., renameVm: true and/or moveToFolder: true), the controller encounters reconciliation failures after the first few successful migrations. Specifically:

After a VM migration succeeds and post-migration rename is applied (e.g., appending _migrated_to_pcd to the VM name in vCenter), the VMware discovery controller detects the change, deletes the old VMwareMachine CR (named based on the original VM name + hash, e.g., centos-7-automation-54c4c), and creates a new one for the renamed VM.
During subsequent reconciles (triggered by later VM completions), the validateMigrationPlanVMs function attempts to validate all VMs in migrationplan.Spec.VirtualMachines (including already-succeeded and renamed ones). This fails with "VMwareMachine not found" errors for the deleted CRs.

As a result:
The entire plan is marked as Failed with a validation error message (e.g., "failed to get VMwareMachine for VM Centos-7-automation: VMwareMachine centos-7-automation-54c4c not found").
Remaining non-terminal migrations (e.g., the 3rd or 4th VM in a 4-VM plan) are incorrectly marked as ValidationFailed, even if their pods complete successfully (as seen in pod logs).
Post-migration actions are not applied to all VMs, as the plan failure skips further reconciliation. Only the first 1-2 VMs get actions before enough renames trigger CR deletions and failures.

Additionally, if there's a timing issue (e.g., vCenter sync lag), post-action operations like rename or move may panic or fail hard if the VM is temporarily "not found" in vCenter.

This leads to incomplete migrations, bogus failure states, and manual intervention required (e.g., deleting failed Migration CRs to retry). Reproducible with 3+ VMs in a plan and rename enabled.

Expected behavior:
Validation should only check remaining non-terminal VMs (pending migrations).
Post-migration actions should apply to all successful VMs without failures propagating from missing/renamed VMs/CRs.
Graceful handling of "not found" errors during post-actions to avoid panics and allow skips (e.g., if discovery interfered).
No plan-wide failure mid-process; all VMs should complete independently.

sarika-pf9:

- Targeted Validation: Only validates non-terminal (pending) VMs, preventing "not found" errors for already migrated or renamed VMs.
- Resilient Post-Actions: Adds "not found" error handling in renameVM and moveVMToFolder to skip already-processed VMs gracefully.
- Panic Prevention: Adds nil checks for Datacenter objects to ensure stable folder auto-detection.
- Plan Stability: Ensures post-migration actions (rename/move) don't trigger "ValidationFailed" or block remaining VMs in the plan.

fixes #1556 

## Testing done

image

### DIFF
--- a/k8s/migration/internal/controller/migrationplan_controller.go
+++ b/k8s/migration/internal/controller/migrationplan_controller.go
@@ -2078,7 +2078,8 @@ func (r *MigrationPlanReconciler) validateMigrationPlanVMs(
 	vmsToValidate []string,
 ) ([]*vjailbreakv1alpha1.VMwareMachine, []*vjailbreakv1alpha1.VMwareMachine, error) {
 	if len(vmsToValidate) == 0 {
-		return nil, nil, fmt.Errorf("no VMs to migrate in migration plan")
+		r.ctxlog.Info("No VMs left to validate; all in terminal states or plan complete")
+		return nil, nil, nil
 	}
 
 	validVMs := make([]*vjailbreakv1alpha1.VMwareMachine, 0, len(vmsToValidate))


k8s/migration/internal/controller/migrationplan_controller.go
```diff
@@ -2067,31 +2075,31 @@ func (r *MigrationPlanReconciler) validateMigrationPlanVMs(
 	migrationplan *vjailbreakv1alpha1.MigrationPlan,
 	migrationtemplate *vjailbreakv1alpha1.MigrationTemplate,
 	vmwcreds *vjailbreakv1alpha1.VMwareCreds,
+	vmsToValidate []string,
 ) ([]*vjailbreakv1alpha1.VMwareMachine, []*vjailbreakv1alpha1.VMwareMachine, error) {
-	var validVMs, skippedVMs []*vjailbreakv1alpha1.VMwareMachine
-
-	if len(migrationplan.Spec.VirtualMachines) == 0 {
+	if len(vmsToValidate) == 0 {
 		return nil, nil, fmt.Errorf("no VMs to migrate in migration plan")
```
OmkarDeshpande7: Please log it correctly. There might be VMs, but they are in terminal state

OmkarDeshpande7:
Also, please link a GHI

PR #1585, opened 2026-02-26
fix: Migration phase skips CopyingChangedBlocks/ConvertingDisk after admin cutover

### LINKED ISSUE #1501: VMs go into emergency mode saying disk not found

Some VM migrations are successful, but the target VM is booting into emergency mode with below error

Image

above issue was found on VM with Centos7

Migrated VM to boot successfully

### Environment

v0.4.0

### vCenter version

any

_No response_

spai-p9:
After triggering admin cutover, the migration status would intermittently skip intermediate phases (CopyingChangedBlocks, ConvertingDisk) and jump directly from AwaitingAdminCutOver to Succeeded. If due to some timing issue, if the below code is not hit:

	if pod.Labels["startCutover"] != "yes" {
				scope.Migration.Status.Phase = vjailbreakv1alpha1.VMMigrationPhaseAwaitingAdminCutOver
				break loop
			}
			// Admin cutover was triggered, reset to a lower phase so it can progress normally
			scope.Migration.Status.Phase = vjailbreakv1alpha1.VMMigrationPhaseCopying

			// If startCutover is "yes", don't set phase here - let it progress to next phase
			// by continuing to check other events
			
The issue occurs when the following case condition in is not satisfied: 
case strings.Contains(events.Items[i].Message, openstackconst.EventMessageWaitingForAdminCutOver) &&
    constants.VMMigrationStatesEnum[scope.Migration.Status.Phase] <= constants.VMMigrationStatesEnum[vjailbreakv1alpha1.VMMigrationPhaseAwaitingAdminCutOver]:

If the current phase is already AwaitingAdminCutOver (15), the condition 15 <= 15 is true and the case is entered. However, if for any reason this case is missed during reconciliation (e.g., the event is not processed or the condition evaluates differently), the phase is never reset to Copying:Without this reset, the phase remains at AwaitingAdminCutOver (15). When subsequent events like "Copying changed blocks" or "Converting disk" are emitted, the controller cannot transition to these phases due to the enum ordering. 

fixes #1501 

## Testing done


OmkarDeshpande7 (approved):
LGTM, good find

PR #1634, opened 2026-03-05
fix - Static network interface name is not getting preserved for Rocky9

### LINKED ISSUE #1620: Static network interface name is not getting preserved for Rocky9

Static network interface name is not getting preserved for Rocky9 -

Image

Static network interfaces should be preserved

### Environment

main

### vCenter version

NA

_No response_

meghansh-pf9:

Improved network configuration detection for RHEL-based systems to avoid overwriting existing static interface configurations.

Previously, the script checked only a single configuration path and would generate DHCP configuration if a static configuration was not found there, even when it existed in other valid locations. The logic now scans all supported paths before making changes.

Additionally, support for generating systemd .link files has been added for interface renaming, since some RHEL versions no longer support legacy udev-based renaming.

fixes #1620

## Testing done

image

image

### DIFF
--- a/scripts/generate-udev-mapping.sh
+++ b/scripts/generate-udev-mapping.sh
@@ -127,7 +127,7 @@ process_rhel(){
   
   local VJB_INDEX=1
   cat "$NET_MAPPING_DATA" | while read -r line_entry; do
-  local found=0
+    local found=0
     parse_address_pair "$line_entry"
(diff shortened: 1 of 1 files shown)


scripts/generate-udev-mapping.sh
sarika-pf9: this too

sarika-pf9 (approved):
lgtm

PR #1708, opened 2026-03-17
feat: Add KVM module loader DaemonSet to enable hardware acceleration…

### LINKED ISSUE #1646: Mount /dev/kvm into the v2v-helper pod so KVM can actually be used

Mount /dev/kvm into the v2v-helper pod so KVM can actually be used

there is no /dev/kvm

qemu-kvm: Could not access KVM kernel module: No such file or directory
qemu-kvm: failed to initialize kvm: No such file or directory
qemu-kvm: falling back to tcg
\x1bc\x1b[?7l\x1b[2J\x1b[0mSeaBIOS (version 1.17.0-8.fc42)
iPXE (https://ipxe.org/) 00:04.0 C000 PCI2.10 PnP PMM+9EFCA9D0+9EF0A9D0 C000
Press Ctrl-B to configure iPXE (PCI 00:04.0)...Sanya Varghese  [1:02 PM]
libguestfs: qemu maxrss 1172164K

~1.1GB

### Detailed Description

Mount /dev/kvm into the v2v-helper pod so KVM can actually be used

_No response_

sanya-pf9:
The v2v-helper pod uses QEMU for disk conversion. Without /dev/kvm, QEMU falls back to software emulation (TCG), which is significantly slower and memory-hungry (observed ~1.1GB RSS vs hardware-accelerated baseline).

This adds a privileged DaemonSet (kvm-module-loader) to the migration-system namespace that loads the kvm-intel or kvm-amd kernel module on the vjailbreak node at startup. Once loaded, /dev/kvm is visible inside the v2v-helper pod via the existing /dev bind mount, enabling QEMU to use KVM hardware acceleration automatically.

fixes #1646

## Special notes for your reviewer
I did my pros and cons list before choosing a daemon set, listing them here 

Design decisions

**Why a DaemonSet and not an init container in the v2v-helper job?**
The v2v-helper runs thousands of migrations on a single node. An init container would run modprobe on every single migration, adding overhead to each one for something that only needs to happen once per
node boot. The DaemonSet loads the module once at startup and it stays loaded for all subsequent migrations.

Why not both (DaemonSet + init container as a safety net)?
modprobe is idempotent, if the module is already loaded it exits immediately. So the init container wouldn't be harmful, but it would be redundant overhead on every migration. The race window where the
DaemonSet hasn't run yet but a migration job has already started is practically unreachable in normal usage (module loads in ~100ms; by the time a user creates a MigrationPlan and the job pod is scheduled, the DaemonSet has already run). Even if the race did occur, QEMU falls back to TCG gracefully rather than crashing.

**Why is the DaemonSet always running (sleep infinity)?**
DaemonSet pods require restartPolicy: Always. The sleep infinity keeps the pod alive after modprobe completes so the DaemonSet considers the node configured. Resource overhead is negligible (~1MB RSS, 0 CPU).

**Minimal RBAC — why no ServiceAccount or RBAC rules?**
The pod never calls the Kubernetes API. automountServiceAccountToken: false ensures no token is mounted. No ClusterRole or RoleBinding is needed. The only elevated privilege is privileged: true on the container (required to load kernel modules) and a read-only /lib/modules host mount (required by modprobe to locate module files).

**Why scoped to control-plane nodes only?**
The vjailbreak node runs with the node-role.kubernetes.io/control-plane label. Scoping the DaemonSet to this label ensures the module loader only runs on the vjailbreak node and not on any other nodes
  in a broader cluster. The toleration for the control-plane:NoSchedule taint is required because control-plane nodes typically carry this taint.

## Testing done

### DIFF
--- a/k8s/migration/internal/controller/esxisshcreds_controller.go
+++ b/k8s/migration/internal/controller/esxisshcreds_controller.go
@@ -23,6 +23,8 @@ import (
 	"sync"
 	"time"
 
+	"golang.org/x/crypto/ssh"
+
 	"github.com/pkg/errors"
 	vjailbreakv1alpha1 "github.com/platform9/vjailbreak/k8s/migration/api/v1alpha1"
 	"github.com/platform9/vjailbreak/pkg/common/constants"
@@ -273,6 +275,10 @@ func (r *ESXiSSHCredsReconciler) getSSHCredentialsFromSecret(ctx context.Context
 		username = "root"
 	}
 
+	if _, err := ssh.ParsePrivateKey(privateKey); err != nil {
+		return nil, fmt.Errorf("invalid SSH private key in secret %s/%s: %v", secretNamespace, secretName, err)
+	}
+
 	return &vjailbreakv1alpha1.ESXiSSHCredsInfo{
 		Username:   username,
 		PrivateKey: privateKey,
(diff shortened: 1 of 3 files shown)


spai-p9:
Closing this since decided to go ahead with doc.

PR #1719, opened 2026-03-18
Added user assigned ip per nic in backend

meghansh-pf9:

Added user assigned ip per nic to enable multiple ips to be assigned by the user


sarika-pf9 (approved):
lgtm

PR #1730, opened 2026-03-20
Feat/support virtual portgroups

### LINKED ISSUE #1717: Orphaned Network shown for DVS port group networks in VMwareMachine CR

When vjailbreak discovers VMs from vCenter, it builds the list of networks for each VM by looking up the network name from vCenter's property collector using a Managed Object Reference (MOR).

The lookup at credutils.go:1728 hardcodes Type: "Network" for all NIC backings:
netRef := types.ManagedObjectReference{Type: "Network", Value: nic.Network}

However, VMware has three distinct network backing types with different MOR types:

Network, DistributedVirtualPortgroup, OpaqueNetwork

When a VM's NIC is connected to a DVS port group (e.g. vlan110 under dvSwitch-DST01), ExtractVirtualNICs correctly extracts the portgroup key (dvportgroup-XXXX) but the subsequent lookup uses the wrong MOR type. vCenter cannot find the object → lookup fails → fallback to the hardcoded string "Orphaned Network" (line 1733).

Fix: Store the MOR type alongside the network value when extracting NIC backing info in ExtractVirtualNICs, and use the correct type when performing the property collector lookup.

### Environment

v4.0

### vCenter version

N/A

_No response_

sanya-pf9:

The current vjb flow does not have a fall back for Distributed Virtual Port Groups/ Opaque Networks.
The lookup at credutils.go:1728 hardcodes Type: "Network" for all NIC backings:
netRef := types.ManagedObjectReference{Type: "Network", Value: nic.Network}

However, VMware has three distinct network backing types with different MOR types:

Network, DistributedVirtualPortgroup, OpaqueNetwork

Fix: Store the MOR type alongside the network value when extracting NIC backing info in ExtractVirtualNICs, and use the correct type when performing the property collector lookup.

fixes #1717

### DIFF
--- a/k8s/migration/pkg/utils/credutils.go
+++ b/k8s/migration/pkg/utils/credutils.go
@@ -821,10 +821,10 @@ func ExtractVirtualNICs(vmProps *mo.VirtualMachine) ([]vjailbreakv1alpha1.NIC, e
 				}
 			case *types.VirtualEthernetCardDistributedVirtualPortBackingInfo:
 				network = backing.Port.PortgroupKey
-				networkType = "DistributedVirtualPortgroup"
+				networkType = constants.VMwareNetworkTypeDistributedVirtualPortgroup
 			case *types.VirtualEthernetCardOpaqueNetworkBackingInfo:
 				network = backing.OpaqueNetworkId
-				networkType = "OpaqueNetwork"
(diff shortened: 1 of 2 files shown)


spai-p9 (commented):
rest looks good, any chance we can test this in our env?

PR #1788, opened 2026-04-06
feat: detect stale Job by owner UID, delete it, fall through to create

### LINKED ISSUE #1787: Retry leaves migration permanently stuck -> stale Job from previous attempt is never replaced

When a migration is retried (via the UI retry button, or by deleting the Migration CR), the MigrationPlan controller silently skips Job creation because a Job with the same name already exists from the previous attempt. No  pod is ever created, and the Migration loops forever with "Migration pod not found yet, requeuing" every 30 seconds.

In CreateJob, the code that checks for an existing Job has an && logic bug:
  err = r.Get(ctx, ..., job)
  if err != nil && apierrors.IsNotFound(err) {   // only enters block when NOT found
      // create job
  }
  // if job already exists (err == nil), falls through and returns nil silently

Why retry worked before 4.1 and broke after? PR #1544

 Before PR #1544, deleting a Migration set a DeletionTimestamp but the Migration wasn't immediately gone (it has a finalizer). CreateMigration got an AlreadyExists error and the reconcile requeued with backoff. During that backoff window, the finalizer was cleared, the Migration was fully deleted, GC deleted the old Job, and by the time the reconcile retried, CreateJob ran against a clean slate.

PR #1544 removed the DeleteFunc watch and made new Migrations start with Phase=Pending. Now the MigrationPlan reconcile is triggered by CreateFunc on the new Migration — which fires immediately after the old Migration is gone, before GC has had a chance to delete the old Job. The stale Job is found, creation is silently skipped, and the migration is stuck permanently.

In CreateJob: when a Job with the expected name already exists, delete it and proceed to create a fresh one.

### Environment

pov env

### vCenter version

N/A

_No response_

sanya-pf9:
In CreateJob: when a Job with the expected name already exists, delete it and proceed to create a fresh one.

fixes #1787

## Testing done
a lot of changes came up in formatting! Please ignore that
Difficult to recreate naturally, it is a race condition, but has occurred thrice. I have mentioned the same on Slack.

### DIFF
--- a/k8s/migration/internal/controller/migrationplan_controller.go
+++ b/k8s/migration/internal/controller/migrationplan_controller.go
@@ -1087,184 +1087,198 @@ func (r *MigrationPlanReconciler) CreateJob(ctx context.Context,
 
 	job := &batchv1.Job{}
 	err = r.Get(ctx, types.NamespacedName{Name: jobName, Namespace: migrationplan.Namespace}, job)
-	if err != nil && apierrors.IsNotFound(err) {
-		r.ctxlog.Info(fmt.Sprintf("Creating new Job '%s' for VM '%s'", jobName, vm))
-		job = &batchv1.Job{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      jobName,
-				Namespace: migrationplan.Namespace,
-			},
-			Spec: batchv1.JobSpec{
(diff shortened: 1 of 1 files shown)


OmkarDeshpande7:
This change is not needed as of now as this is not reproduced. Also the original issue was caused due to mis-configuration for the VJB vm. Will re-open if needed.

PR #1805, opened 2026-04-10
fix - Interface preservation not working on DHCP disabled network for windows

### LINKED ISSUE #1786: Windows static IP / Interface name preservation does not work when DHCP is disabled on destination network

Windows static IP / Interface name preservation does not work when DHCP is disabled on destination network

Image

Image

Note: Port IP is same as source IP but issue is with IP inside the VM.

IP and interface name should be preserved.

### Environment

main

### vCenter version

NA

_No response_

meghansh-pf9:

Added offline mechanisms to recover and restore ip interface configurations

fixes #1786 

## DHCP disabled on network
image


OmkarDeshpande7 (approved):
LGTM

PR #1834, opened 2026-04-17
feat: add volume image properties prior to migration

### LINKED ISSUE #931: RFE - provision of an option the migration form to set the PCD volume metadata
### Issue:

### Requirement

_No response_

### LINKED ISSUE #1749: Adjusting Image Properties Prior to Migration

Feature that allows to add or modify "Image Properties" (volume_image_metadata) on the volumes before the Instance is built on destination (PCD) – to avoid rebuilds or "Hard Reboots“.
For example like Adding "os_require_quiesce=no“. image property to the boot volume of the VM during the migration – or modify "hw_firmware_type=uefi“ to "Bios“ 

### Detailed Description

allow modifying / setting the volume_image_metadata both existing parameters and new volume_image_metadata parameters to the VM's volumes
This should work from UI and vjb API

_No response_

sarika-pf9:
feat: add image properties prior to migration

fixes #1749 
fixes #931 

## Testing done

image

image

image


spai-p9 (approved):
Looks good

PR #1841, opened 2026-04-22
UI: show selected profiles in details modal

### LINKED ISSUE #931: RFE - provision of an option the migration form to set the PCD volume metadata

### LINKED ISSUE #1843: Show selected image profiles in migration details modal
Show selected image profiles in migration details modal to keep track from the UI

sarika-pf9:
UI: show selected profiles in details modal

fixes #1843 
fixes #931 

## Testing done
image
image


AbhijeetThakur (approved):
LGTM

PR #1842, opened 2026-04-22
Handling interfaces with no ip for rhel based os

### LINKED ISSUE #1832: Centos7 - Destination VM gets IP for an interface which has no IP on source

Persist network interface option is selected on the source.
Image
Image

Should preserve only source interfaces

### Environment

main

### vCenter version

NA

_No response_

meghansh-pf9:

Explicitly injecting a no ip configuration for interfaces having no ip at the source to maintain consistency

fixes #1832 

### No ip configuration being injected during migration 
image

### Configuration at destination with no ip 
image

### DIFF
--- a/scripts/generate-udev-mapping.sh
+++ b/scripts/generate-udev-mapping.sh
@@ -135,7 +135,31 @@ process_rhel(){
     fi
     
     if [[ -z "$FOUND_IP" ]]; then
-      display_msg "Skipping entry with empty IP for MAC $FOUND_MAC: $line_entry"
+      display_msg "No IP found for MAC $FOUND_MAC, generating no-IP interface: $line_entry"
+      if [[ -n "$TARGET_DIR" && -d "$TARGET_DIR" ]]; then
+        {
+          echo "TYPE=Ethernet"
+          echo "BOOTPROTO=none"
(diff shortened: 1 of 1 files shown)


scripts/generate-udev-mapping.sh
```diff
@@ -135,7 +135,31 @@ process_rhel(){
...
+      elif [[ -d "$NM_CONN_PATH" ]]; then
+        {
+          echo "[connection]"
+          echo "id=vjb$VJB_INDEX"
+          echo "type=ethernet"
+          echo "interface-name=vjb$VJB_INDEX"
+          echo "mac-address=$FOUND_MAC"
+          echo "[ipv4]"
+          echo "method=disabled"
+        } > "$NM_CONN_PATH/vjb$VJB_INDEX.nmconnection"
```
sarika-pf9, line 160: Missing [ipv6] section in the .nmconnection file

For RHEL 8/9 NM connections, omitting [ipv6] leaves IPv6 in default (often auto) mode. If the source interface truly has no IP, the destination should also disable IPv6:

[ipv6]
method=disabled
The existing fallback at line 293–303 also omits this — so this is a pre-existing gap, but worth fixing here

sarika-pf9 (approved):
looks good

PR #1846, opened 2026-04-22
Remove default profile auto-selection and fix boot volume metadata merge

### LINKED ISSUE #931: RFE - provision of an option the migration form to set the PCD volume metadata

### LINKED ISSUE #1844: Remove default profile selection during migration

Remove default profile selection during migration

We found that, a bios boot VM also getting uefi property set by default, causing the boot issue -

Image
Image

By default if user has not selected any profiles, we should fallback to default properties we set in the code. Only if user selects a profile and there is a conflict in default and user properties, we should honour the user properties.

### Environment

main

### vCenter version

NA

_No response_

sarika-pf9:

- Remove auto-selection of default-windows/default-linux profiles in migration form; profiles are now opt-in only
- Merge profile properties on top of existing metadata, conflicting keys resolve to the user's profile value
- Align default profile YAMLs to exactly match hardcoded properties (hw_qemu_guest_agent, hw_video_model, hw_pointer_model for both OS families; hw_disk_bus, os_type additionally for Windows)

fixes #1844 
fixes #931 

## Testing done

- no auto selection
image

- confict on same properties with different key value
image

### Test Case 1
- selected two profiles with diff values:
image

- it got migrated with union of all propperties
image

### Test Case 2

- selected 3 different profiles with different values and kept one value different for default property to override
image

- it got migrated with union of all and override the value of default `hw_qemu_guest_agent` and detected source VM is UEFI so auto applied `hw_firmware_type=uefi`
image

### Test Case 3

- selected no profile 
image

- it got migrated with default image properties and detected source VM is UEFI so auto applied `hw_firmware_type=uefi`

image

### Test Case 4

- selected profile with an addition property `os_type`
image

- it got migrated with proper union of all properties

image

## didn't touched image properties for data devices of any VM
image

### DIFF
--- a/deploy/volumeimageprofile-defaults.yaml
+++ b/deploy/volumeimageprofile-defaults.yaml
@@ -7,11 +7,11 @@ spec:
   osFamily: windowsGuest
   description: Default profile for Windows VMs
   properties:
-    hw_disk_bus: virtio
-    hw_vif_model: virtio
-    os_type: windows
     hw_qemu_guest_agent: "yes"
-    hw_video_model: qxl
+    hw_video_model: virtio
+    hw_pointer_model: usbtablet
+    hw_disk_bus: virtio
+    os_type: windows
 ---
 apiVersion: vjailbreak.k8s.pf9.io/v1alpha1
 kind: VolumeImageProfile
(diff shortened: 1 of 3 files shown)


OmkarDeshpande7 (approved):
lgtm

PR #1860, opened 2026-04-27
use custom http timeout from settings

### LINKED ISSUE #1859: Need API timeout to be configurable in vjailbreak settings

If the vjailbreak client timeout can be configured via settings, it will help mitigate this issue in slow-responsive environments.

### Detailed Description

Migrations receive the error "Client.Timeout exceeded" for API calls that actually succeed but take longer than the 30-second timeout threshold. The API status is 200, and the response time can exceed 43 seconds.

Impact
It affects vjailbreak communicating with the OpenStack API through. Calls experience premature timeouts when API calls take longer than 30 seconds, resulting in "Client.Timeout exceeded" errors despite successful API responses.

Expected behaviour
The system should allow API calls to complete successfully without timing out, even if they take longer than 30 seconds.

_No response_

OmkarDeshpande7:

In this pr we add the ability to set custom timeout for apis. in case user needs to adjust the timeouts based on their network delays

fixes #1859 

### Http Timeout in UI 
image


meghansh-pf9 (approved):
lgtm

AbhijeetThakur (approved):

PR #1866, opened 2026-04-27
Map lun to igroup based on svm

### LINKED ISSUE #1864: Bug: NetApp XCOPY migration fails with 404 "No such initiator group exists" when a same-named igroup exists on another SVM

StorageAcceleratedCopy (XCOPY) migrations to NetApp ONTAP fail at the LUN-mapping step with an ONTAP 404 No such initiator group exists error. The failure happens whenever the ONTAP cluster has an igroup on a different SVM that contains the same FC WWPNs (or iSCSI IQNs) as the ESXi host.                                                             
  

Migration to get successfully completed.

### Environment

v0.4.4-rc

### vCenter version

NA

NA

spai-p9:
There was a bug where if there is a igroup with the esxi wwpn we would pick that irrespective of the SVM, and when we try to map the volume which is present on SVM-B it would not let us map the igroup we found with the wwpn of the esxi on SVM-A. Hence for this logic we should do a svm based filtering and map the volume to the svm given by the end user. 

fixes #1864 

## Testing done

### DIFF
--- a/pkg/vpwned/sdk/storage/netapp/netapp.go
+++ b/pkg/vpwned/sdk/storage/netapp/netapp.go
@@ -389,14 +389,15 @@ func normaliseToONTAPInitiators(hbaIdentifiers []string) ([]string, error) {
 
 // ensureIgroupExists returns the UUID of the named igroup on svmName, creating it
 // (with the given protocol and os_type "vmware") if it does not already exist.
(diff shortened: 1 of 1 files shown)


OmkarDeshpande7 (approved):
lgtm

PR #1892, opened 2026-05-04
[1241] Disable port security on migrated ports when no security group is selected while migration

### LINKED ISSUE #1241: When security group is not selected it causes traffic drops

when security group is not selected in vjb (and security is enabled in network level), it enables the port security without selecting a security group which causes traffic drops post migration

The normal behaviour is when we do not specify the security group, pcd picks the default one automatically. but it is not the case with vjb, it is enabling the security group on the port but without choosing a security group which is causing the traffic to be dropped 

issue is caused by changes on pcd network recently. we have enabled the security group on all networks

Image

Image

The normal behaviour is when we do not specify the security group, pcd picks the default one automatically. but it is not the case with vjb, it is enabling the security group on the port but without choosing a security group which is causing the traffic to be dropped 

### Environment

pcd Oct release
vjb 0.3.6 and 0.3.4

### vCenter version

8.

_No response_

sarika-pf9:
When a user migrates a VM without selecting a security group, vjailbreak was creating the port with security_groups: [] — an empty list. Since PCD recently enabled port security on all networks, OpenStack interpreted this as "port security enabled, no rules," causing all traffic to be dropped post-migration.

**Fix:** When no security groups are selected, the port is now explicitly created with port_security_enabled: false. 

fixes #1241 

## Testing done

- Selected no security group while creating migration
image

- Selected one while creating migration
 
image

### DIFF
--- a/v2v-helper/pkg/utils/openstackopsutils.go
+++ b/v2v-helper/pkg/utils/openstackopsutils.go
@@ -35,6 +35,7 @@ import (
 	"github.com/gophercloud/gophercloud/v2/openstack/compute/v2/servers"
 	"github.com/gophercloud/gophercloud/v2/openstack/compute/v2/volumeattach"
 	"github.com/gophercloud/gophercloud/v2/openstack/identity/v3/projects"
+	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/portsecurity"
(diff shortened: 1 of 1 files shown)


spai-p9 (approved):
looks good.

PR #1896, opened 2026-05-04
[1897] Feat: Add controller logs button to topbar in UI

### LINKED ISSUE #1897: View controller logs without triggering a migration
Currently, controller logs are only accessible through the per-migration logs drawer, which means a user must first create and trigger a migration to see any controller activity. There is no way to inspect the controller manager logs when the system is idle or when debugging issues unrelated to a specific migration.

Expectation: Expose controller elsewhere so they are always accessible regardless of migration state.

sarika-pf9:
- Adds a terminal icon button to the topbar that opens a Controller Logs drawer, accessible from any page without needing an active migration
- Removes the Controller tab from the per-migration LogsDrawer — migration pod logs and controller logs are now cleanly separated: migration-specific logs stay in the migrations table drawer, controller-manager logs are always available globally

fixes #1897 

## Testing done
image
image
image


AbhijeetThakur (approved):
LGTM

PR #1903, opened 2026-05-06
[1907] Fix: fetch single VMwareMachine instead of entire list in migration detail query

### LINKED ISSUE #1907: Bug: migration details modal for a single vm do query for entire vmwaremachines
Bug: migration details modal for a single vm do query for entire vmwaremachines. we should query only for the vm we are opening the modal for.

sarika-pf9:
- Previously, the migration detail modal fetched all VMwareMachine resources and filtered in-memory to find the relevant one. Since migration names are always migration-<vmwareMachineName>, we can derive the exact resource name and fetch only that single VMwareMachine directly via the Kubernetes API.

- Also added ability to multiselect profiles at once

fixes #1907 

## Testing done

image


AbhijeetThakur (approved):
LGTM

PR #1910, opened 2026-05-06
[1830] Fix: Include RBAC-shared security groups in destination tenant security group list

### LINKED ISSUE #1830: vjailbreak does not show shared security groups in target/destination tenant

- In a customer environment, security groups are managed in a central tenant and shared with all other tenants via Neutron RBAC (access_as_shared).
- Observation: When attempting a migration in vjailbreak for these tenants, the shared security groups are missing from the destination tenant's security group list.
- Shared security groups are visible and usable via the PCD UI but vjailbreak fails to discover them, showing only the security groups natively owned by the tenant.

I expected vjailbreak to have parity with the PCD/OpenStack. It should list all security groups available to the destination tenant (both owned and RBAC-shared) so that migrations can be mapped to the customer's required security standards without needing external post-migration scripts.

### Environment

vjailbreak: 0.4.3

### vCenter version

Version: 8.0.3

_No response_

sarika-pf9:
Security groups shared via Neutron RBAC (`access_as_shared`) were not appearing in the destination tenant's security group dropdown in vJailbreak.

**Fix:**
- Added support for fetching RBAC-shared security groups (including those shared with `*` or specific projects).
- Updated security group listing logic to show both owned and accessible shared security groups.

Fix:

fixes #1830 

## Testing done
image
image
image
image
image


OmkarDeshpande7 (approved):
lgtm

PR #1935, opened 2026-05-13
[1888] refactor: Remove direct pod permissions

### LINKED ISSUE #1888: Reduce RBAC scope for migration role by delegating secrets and pod access via proxy
The current ui-manager-role ClusterRole has direct permissions to access sensitive resources such as secrets and pods. 

Expectation:
Reduced RBAC permissions for ui-manager-role.

sarika-pf9:

**Before:** ui-manager-sa had direct Kubernetes RBAC for secrets (full CRUD) and pods (get/list/watch/patch). Any compromise of this service account could directly read/modify secrets and pods.

**After:** Those RBAC rules are removed from ui-manager-role. The UI still works because all secret and pod operations are routed through the vpwned API server (running as migration-controller-manager) via a reverse proxy at /vpw/v1/k8s/*. A compromised ui-manager-sa token has no direct K8s API access for secrets or pods.

### Two layers enforced before any request is forwarded:

- Layer 1 — TokenReview gate: vpwned calls the K8s TokenReview API to verify the incoming Authorization: Bearer token is (a) cryptographically valid and (b) belongs specifically to system:serviceaccount:migration-system:ui-manager-sa. Invalid or missing tokens → 401. Valid tokens cached 60 seconds.

- Layer 2 — Path + method allowlist: Even with a valid token, only the exact set of K8s API calls the UI makes are allowed through. Everything else → 403.

Resource | Methods allowed
-- | --
pods | GET (list/get/log), PATCH (admin cutover)
secrets | GET, POST, PUT, DELETE
anything else | blocked

fixes #1888

## Testing done
did migrations with admin initiated cutover from the fresh build
image

### DIFF
--- a/deploy/00crds.yaml
+++ b/deploy/00crds.yaml
@@ -4672,17 +4672,6 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - secrets
-  verbs:
-  - create
-  - delete
-  - get
-  - list
-  - patch
-  - update
-- apiGroups:
-  - ""
-  resources:
   - configmaps
   verbs:
   - create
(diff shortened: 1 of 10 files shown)


OmkarDeshpande7 (approved):
lgtm

PR #1936, opened 2026-05-14
[1938] Support VMs with space in their name

### LINKED ISSUE #1938: Bug: VMs with spaces in their name fail to migrate
Unable to create Migration CR for VMs that have spaces in their name.

Expectation
Migration should proceed for VMs having space in their name

sarika-pf9:

**Problem:**
VMs with spaces in their names (e.g. "sarika - 2k19") could not be migrated. The controller would fail at Migration CR creation with:

```
metadata.labels: Invalid value: "sarika - 2k19":
a valid label must be an empty string or consist of alphanumeric characters,
'-', '_' or '.', and must start and end with an alphanumeric character
```

**Root Cause:**
we have a label `vjailbreak.k8s.pf9.io/vm-key` on Migration CRs to store the VM's unique key (vmName-moid format). The raw key was written directly as the label value. Kubernetes label values do not allow spaces or most special characters, so any VM whose name contained spaces caused the API server to reject the Migration CR entirely.

`Fix:`
The label is kept but its value is now sanitized to satisfy Kubernetes constraints — this sanitized value is used only as an internal index and is never used to identify or reference the VM in any operation. The original VM name, exactly as entered — including spaces, special characters, and full length — is preserved in a new annotation `vjailbreak.k8s.pf9.io/original-vm-name`. All functional lookups  read the annotation to get the original name. 

fixes #1938 

## Testing done
image
image

### DIFF
--- a/k8s/migration/api/v1alpha1/migration_types.go
+++ b/k8s/migration/api/v1alpha1/migration_types.go
@@ -86,9 +86,10 @@ type MigrationSpec struct {
 	// VMName is the name of the VM getting migrated from VMWare to Openstack
 	VMName string `json:"vmName"`
 
-	// DisplayName is the original VM display name including any special characters (spaces, etc.).
-	// VMName holds the same value; this field ensures the original name is accessible via spec
-	// even when the label value has been sanitized for Kubernetes compatibility.
+	// DisplayName is the original VM unique key (name-<moid> format) including any special characters
+	// such as spaces. Unlike VMName (which holds only the plain VM name), this field preserves the
+	// full unique key so it remains accessible via spec even when the label value has been sanitized
+	// for Kubernetes compatibility.
 	// +optional
 	DisplayName string `json:"displayName,omitempty"`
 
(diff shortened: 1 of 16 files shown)


k8s/migration/api/v1alpha1/migration_types.go
```diff
@@ -86,6 +86,13 @@ type MigrationSpec struct {
 	// VMName is the name of the VM getting migrated from VMWare to Openstack
 	VMName string `json:"vmName"`
 
+	// DisplayName is the original VM unique key (name-<moid> format) including any special characters
+	// such as spaces. Unlike VMName (which holds only the plain VM name), this field preserves the
+	// full unique key so it remains accessible via spec even when the label value has been sanitized
+	// for Kubernetes compatibility.
+	// +optional
+	DisplayName string `json:"displayName,omitempty"`
```
OmkarDeshpande7: cannot we construct it every time?

OmkarDeshpande7 (commented):
@sarika-pf9 please add RCA and summary of the changes performed in the PR description. from the changes, I am unable to get whet you are trying to do

k8s/migration/api/v1alpha1/migration_types.go
sarika-pf9: we can

OmkarDeshpande7 (approved):
lgtm

PR #1973, opened 2026-05-27
[1961] fix: Handle transient vCenter errors during VM power-off

### LINKED ISSUE #1961: Add retry logic for guestOS shutdown in vcenter

In vjb 0.4.5, migrations are failing because the VM does not shut down properly in VMware. It throws this error:

```
Guest shutdown failed, falling back to power off: failed to get VM power state: Post "https://xyz-xyz-xy.xyz.abc.com/sdk": context deadline exceeded
2026/05/15 09:10:09 failed to live replicate disks: failed to power off VM: failed to refresh VM reference: VM not found. Trying to perform cleanup
```

If the first call to guestosshutdown fails due to a timeout or connection issue, it should retry the shutdown call instead of immediately failing.

### Environment

VJB 0.4.5

### vCenter version

8.0.3

_No response_

sarika-pf9:

**Bug 1** — `GetVMWithDatacenter` swallows real vCenter errors
Root cause: When `VCFinder.VirtualMachine()` fails with any error (timeout, network, etc.), the loop silently continues to the next datacenter. After exhausting all datacenters it returns "VM not found" — even if the real error was context deadline exceeded. Callers saw a misleading inventory error instead of the actual transport failure.

**Fix:** Only skip to the next datacenter for genuine not-found results. Surface all other errors (timeouts, dropped connections) immediately with the real error message.

**Bug 2** — VMPowerOff fallback uses `GetVMByName` without re-login
Root cause: After guest shutdown fails (e.g. due to HTTP timeout on an overloaded vCenter), the fallback calls `vcclient.GetVMByName` directly. This skips session refresh, so the stale/broken connection from the prior timeout causes `GetVMWithDatacenter `to time out again — which (Bug 1) returns "VM not found". The migration aborts with a false error and the VM is left running.

**Fix:** Replace the raw `GetVMByName` call with `vmops.RefreshVM()`, which re-logins to vCenter before looking up the VM. This is the same pattern used correctly in every other operation.

**Bug 3** — All vCenter ops only retry on NotAuthenticated, ignoring transient network/timeout errors
Root cause: The retry guard throughout `vmops.go` was `strings.Contains(err.Error(), "NotAuthenticated")`. Transient errors like context deadline exceeded, EOF, connection reset, and i/o timeout fell through without retry — causing immediate failure on any brief vCenter overload.

**Fix:** Introduced `IsTransientVCenterError(err)` (exported from the `vcenter` package, reused by both `vcenter` and `vm` packages) that covers:

- NotAuthenticated (session expiry)
- context deadline exceeded (HTTP timeout)
- connection reset by peer
- connection refused
- EOF
- i/o timeout
- no such host

Applied to all vCenter calls.
Additionally fixed the `VMGuestShutdown power-state` polling loop, which had no error handling at all — a transient error mid-wait caused immediate failure instead of retrying.

fixes #1961

### DIFF
--- a/v2v-helper/vcenter/vcenterops.go
+++ b/v2v-helper/vcenter/vcenterops.go
@@ -160,6 +160,22 @@ func GetThumbprint(host string) (string, error) {
 	return thumbprint, nil
 }
 
+// IsTransientVCenterError returns true for errors that may resolve after refreshing
+// the vCenter session — covers auth expiry, HTTP timeouts, and dropped connections.
+func IsTransientVCenterError(err error) bool {
+	if err == nil {
+		return false
+	}
+	msg := err.Error()
+	return strings.Contains(msg, "NotAuthenticated") ||
+		strings.Contains(msg, "context deadline exceeded") ||
+		strings.Contains(msg, "connection reset") ||
+		strings.Contains(msg, "connection refused") ||
+		strings.Contains(msg, "EOF") ||
+		strings.Contains(msg, "i/o timeout") ||
+		strings.Contains(msg, "no such host")
+}
+
 // Get all datacenters with retry and explicit authentication
 func (vcclient *VCenterClient) getDatacenters(ctx context.Context) ([]*object.Datacenter, error) {
 	// Create a new finder with the current client each time to ensure we're using the most up-to-date client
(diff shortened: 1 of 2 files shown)


OmkarDeshpande7 (approved):
lgtm

PR #2008, opened 2026-06-09
[2009] Unified logs icons and aligned migration name across logs drawer and details

### LINKED ISSUE #2009: Migration resource name mismatch between details and logs drawer
Two UI consistency issues on the Migrations page:

- Icons — the controller-logs button (Terminal icon) and the pod-logs button (ListAlt icon) use different icons for the same kind of action (viewing logs).

Image

- Name — for the same migration, the Migration Details modal shows the VM name while the logs drawer shows a derived object name. The two should display the same name.

Image

Image

sarika-pf9:
- Replaced the controller-logs Terminal icon with the ListAlt icon so it matches the pod-logs icon — both log entry points now look identical.
- The migration logs drawer header now shows the migration's spec.vmName (same as the Migration Details modal) instead of a derived object name, keeping the resource name consistent across both views.

fixes #2009 

## Testing done

image

image
image


AbhijeetThakur (approved):
LGTM

PR #2015, opened 2026-06-10
Fixed the missing --binding-profile '{"l2-port": true}' on ports created for VMs migrated onto L2-only networks in PCD

### LINKED ISSUE #2014: L2 port missing binding profile configuration

Customer migrated a SQL server workload using vJailbreak (rel -0.4,5) and migrated VM on PCD didn't have the tag  --binding-profile '{"l2-port": true}' assigned to the port. 
This leads to the failure of the workload VM when the team uses Veeam to restore a backed up VM with the following error

Image

**Steps taken**
After updating the port with the correct binding profile:

openstack port set --binding-profile '{"l2-port": true}' <port_id>

inside the VM appliance, resolves the issue since the PCD Veeam Proxy uses that binding profile to check if it's an L2 network or not.

Since the binding profile is not added to VMs migrated with vJailbreak, workloads will not have the tag automatically created and hence the restore fails in those workloads.

As per PCD [docs](https://docs.platform9.com/private-cloud-director/virtualized-networking/layer-2-networking#using-pcdctl-cli-1), the VJB code is missing the --binding-profile '{"l2-port": true}'  when creating a L2 port here in the [code](https://github.com/platform9/vjailbreak/blob/f70f32580b00fdb5ffe9c6020757340d9e748183/v2v-helper/pkg/utils/openstackopsutils.go#L678)

Once this is added, the issue should be fixed from VJB side itself

### Environment

PCD Release - PCD 2026.4-89
vJB version - 0.4.5

### vCenter version

vCenter Server 7.0.3, 24201990
ESXi 7.0.3, 24585291

_No response_

spai-p9:
All workload port creation funnels through a single chokepoint, createPortLowLevel in v2v-helper/pkg/utils/openstackopsutils.go:880 (the only ports.Create call in the repo). It set port security but never attached the L2 binding profile, even though the code already detects L2 networks via the simple_network tag (GetIsSimpleNetwork).

Changes — v2v-helper/pkg/utils/openstackopsutils.go

1. Added the portsbinding import (already part of the in-use gophercloud/v2 v2.9.0 — no go.mod change).
2. Extracted a pure helper buildPortCreateOptions(createOpts, isL2Network) that layers the OpenStack port extensions:
  - L2 network → portsbinding.CreateOptsExt with Profile: {"l2-port": true}
  - No security groups → portsecurity.PortCreateOptsExt with port security disabled
  - Both compose when both conditions apply (each implements ports.CreateOptsBuilder).
3. createPortLowLevel now resolves isL2Network once (via the existing GetIsSimpleNetwork) before the retry loop and uses the helper. A lookup failure is non-fatal — it logs and creates the port without the profile rather than failing the migration.

fixes #2014 

## Testing done

---

<a href="https://app.devin.ai/review/platform9/vjailbreak/pull/2015" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/graphify-out/GRAPH_REPORT.md
+++ b/graphify-out/GRAPH_REPORT.md
@@ -1,286 +1,264 @@
-# Graph Report - vjailbreak  (2026-06-09)
+# Graph Report - vjailbreak  (2026-06-11)
 
 ## Corpus Check
-- 733 files · ~1,732,143 words
+- 419 files · ~616,240 words
 - Verdict: corpus is large enough that graph structure adds value.
 
 ## Summary
-- 5709 nodes · 9401 edges · 749 communities (286 shown, 463 thin omitted)
-- Extraction: 90% EXTRACTED · 10% INFERRED · 0% AMBIGUOUS · INFERRED: 943 edges (avg confidence: 0.8)
+- 5195 nodes · 13962 edges · 259 communities detected
+- Extraction: 48% EXTRACTED · 52% INFERRED · 0% AMBIGUOUS · INFERRED: 7205 edges (avg confidence: 0.8)
 - Token cost: 0 input · 0 output
 
-## Graph Freshness
-- Built from commit: `590cb543`
-- Run `git rev-parse HEAD` and compare to check if the graph is stale.
-- Run `graphify update .` after code changes (no API cost).
-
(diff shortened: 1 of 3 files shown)


geet-pf9 (approved):
LGMT

PR #2024, opened 2026-06-17
[1750] UI for edit and retry 

### LINKED ISSUE #1750: Modifying Existing Migrations in Terms of Flavor, Datastore, or Other Settings

When we create a Migration in the UI and this migration fails or has problem, maybe because of selected options, we would like to have the possibility modify the existing Migration via UI / API / CLI and then retry
At the moment you can only retry failover or delete but not modify (This would lead to a complete copy again)

### Detailed Description

When we create a Migration in the UI and this migration fails or has problem, maybe because of selected options, we would like to have the possibility modify the existing Migration via UI / API / CLI and then retry
At the moment you can only retry failover or delete but not modify (This would lead to a complete copy again)

_No response_

sarika-pf9:
When a migration fails due to misconfiguration (wrong flavor, network/storage mapping, etc.), clicking Retry now opens the migration form pre-populated with the failed VM's existing configuration. Single-VM plans are patched in place; multi-VM plans are cloned into a new 1-VM plan so the retry is isolated and other VMs in the original plan are unaffected.

fixes #1750 

## Testing done

image
image


ui/src/features/migration/hooks/useRetryPrefill.ts
AbhijeetThakur, line 1: Can you refactor this hook? Use React Query for api calls.

ui/src/features/migration/hooks/useRetrySubmit.ts
AbhijeetThakur, line 1: Also, refactor this hook?

AbhijeetThakur (approved):
LGTM

spai-p9 (approved):
lgtm

PR #2034, opened 2026-06-18
[1979] VM selection refresh button now revalidates VMware credentials

### LINKED ISSUE #1979: [UX] Refresh icon in migration form VM table should also re-validate

Currently Refresh icon in VM table in migration form just refresh (re-fetch) the VM list. It should instead, revalidate both creds. Other wise user has to go back to creds page, re-validate, come back to migration form fill form, then migrate.

### Detailed Description

This will save some steps for the user

_No response_

sarika-pf9:
Previously, the refresh icon in the VM selection step only refetched the VM list. This PR makes it also revalidate VMware credentials — identical behavior to the revalidate icon on the credentials page.

fixes #1979 

## Testing done

https://github.com/user-attachments/assets/6b684fd6-88ba-4f8e-b923-0077bbc20fa8

---

<a href="https://app.devin.ai/review/platform9/vjailbreak/pull/2034" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


ui/src/hooks/api/useVmwareRevalidation.ts
```diff
@@ -0,0 +1,144 @@
...
+      return isRevalidating || isBackendRevalidating ? 5000 : false
+    },
+  })
+
+  const { mutate: revalidate, isPending: isRevalidationApiPending } = useMutation({
+    mutationFn: revalidateCredentials,
+    onSuccess: () => {
+      setTimeout(() => {
+        queryClient.invalidateQueries({ queryKey: VMWARE_CREDS_QUERY_KEY })
+      }, 500)
```
AbhijeetThakur, line 77: Why is a fixed 500ms delay required before invalidating the query?

This introduces a timing dependency on backend behavior and may become flaky if backend processing time changes. Consider invalidating/refetching immediately and relying on polling/backend status to determine completion instead of a hardcoded delay.

ui/src/hooks/api/useVmwareRevalidation.ts
```diff
@@ -0,0 +1,144 @@
...
+  onRevalidationComplete?: () => void
+}
+
+export function useVmwareRevalidation({
+  vmwareCredName,
+  onRevalidationComplete,
+}: UseVmwareRevalidationProps) {
+  const { reportError } = useErrorHandler({ component: 'VmsSelectionStep' })
+  const queryClient = useQueryClient()
+  const [isRevalidating, setIsRevalidating] = useState(false)
```
AbhijeetThakur, line 32: We are maintaining revalidation state locally while also deriving revalidation status from the backend (vmwareValidationStatus === 'Revalidating').

Having two sources of truth increases complexity and can lead to state reconciliation issues after refreshes or partial failures. Can we derive the UI state from mutation status + backend status instead?

ui/src/hooks/api/useVmwareRevalidation.ts
```diff
@@ -0,0 +1,144 @@
...
+    vmwareCredName,
+    clearActiveRevalidation,
+    onRevalidationComplete,
+  ])
+
+  useEffect(() => clearRevalidationTimeout, [clearRevalidationTimeout])
+
+  const handleRefreshAndRevalidate = useCallback(() => {
+    if (!vmwareCredName) return
+    const cred = vmwareCreds?.find((c) => c.metadata.name === vmwareCredName)
```
AbhijeetThakur: The same credential lookup is performed multiple times (line numbers 50, 76 & 104)  throughout the hook.

Consider memoizing the selected credential once (e.g., via useMemo) and reusing it across the hook to improve readability and avoid repeated array scans.

ui/src/hooks/api/useVmwareRevalidation.ts
```diff
@@ -0,0 +1,144 @@
...
+    isRevalidating,
+    isRevalidationApiPending,
+    vmwareCredName,
+    clearActiveRevalidation,
+    onRevalidationComplete,
+  ])
+
+  useEffect(() => clearRevalidationTimeout, [clearRevalidationTimeout])
+
+  const handleRefreshAndRevalidate = useCallback(() => {
```
AbhijeetThakur, line 119: Should we prevent revalidation from being triggered while a previous revalidation is already in progress?

Multiple clicks could generate concurrent mutation requests and repeatedly reset the timeout state.

`if (isRevalidating || isRevalidationApiPending) {
  return
}`

ui/src/hooks/api/useVmwareRevalidation.ts
```diff
@@ -0,0 +1,144 @@
...
+    setTimedOutRevalidating(false)
+    setIsRevalidating(true)
+
+    revalidationTimeoutRef.current = setTimeout(() => {
+      revalidationTimeoutRef.current = null
+      setTimedOutRevalidating(true)
+      setIsRevalidating(false)
+      reportError(
+        new Error(
+          'VMware credential revalidation is still in progress after 31 minutes. You can retry.'
```
AbhijeetThakur: The displayed timeout duration is hardcoded while the actual timeout is controlled by REVALIDATION_TIMEOUT_MS.

Consider deriving the displayed duration from the constant to avoid the message becoming inaccurate if the timeout value changes.

ui/src/hooks/api/useVmwareRevalidation.ts
```diff
@@ -0,0 +1,144 @@
...
+  })
+
+  useEffect(() => {
+    if (!isRevalidating || !vmwareCredName) return
+    const targetCred = vmwareCreds?.find((cred) => cred.metadata.name === vmwareCredName)
+    const hasFreshStatus = dataUpdatedAt > revalidationStartDataUpdatedAtRef.current
+
+    if (targetCred) {
+      const status = normalizeStatus(targetCred.status?.vmwareValidationStatus)
+      if (!isRevalidationApiPending && hasFreshStatus && status !== 'Revalidating') {
```
AbhijeetThakur: The completion logic currently combines mutation state, query freshness, and backend status in a single conditional.

Consider extracting this into a well-named helper (e.g. isRevalidationComplete) to improve readability and make future changes safer.

ui/src/hooks/api/useVmwareRevalidation.ts
```diff
@@ -0,0 +1,144 @@
...
+  const clearActiveRevalidation = useCallback(() => {
+    clearRevalidationTimeout()
+    setIsRevalidating(false)
+  }, [clearRevalidationTimeout])
+
+  const { data: vmwareCreds, dataUpdatedAt } = useVmwareCredentialsQuery(undefined, {
+    enabled: !!vmwareCredName,
+    staleTime: 0,
+    refetchOnMount: false,
+    refetchInterval: (query) => {
```
AbhijeetThakur: refetchInterval executes frequently and performs a credential lookup on every invocation.

If the credentials list can grow large, this could become unnecessarily expensive. Long term, it may be worth querying/polling a single credential instead of the entire credentials collection.

ui/src/hooks/api/useVmwareRevalidation.ts
```diff
@@ -0,0 +1,144 @@
...
+        new Error(
+          'VMware credential revalidation is still in progress after 31 minutes. You can retry.'
+        ),
+        {
+          context: 'vmware-revalidation-timeout',
+          metadata: { credentialName: vmwareCredName },
+        }
+      )
+      queryClient.invalidateQueries({ queryKey: VMWARE_CREDS_QUERY_KEY })
+    }, REVALIDATION_TIMEOUT_MS)
```
AbhijeetThakur, line 149: After timeout we invalidate the query, but if the backend still reports Revalidating, polling may resume immediately via refetchInterval.

Is this intentional? If timeout is meant to stop active revalidation tracking, we may want to ensure polling does not automatically restart.

sarika-pf9: will do this too

ui/src/hooks/api/useVmwareRevalidation.ts
sarika-pf9: will take care of this

ui/src/hooks/api/useVmwareRevalidation.ts
sarika-pf9: This is a valid readability suggestion and we can apply it.

ui/src/hooks/api/useVmwareRevalidation.ts
sarika-pf9: will take a look to this

ui/src/hooks/api/useVmwareRevalidation.ts
sarika-pf9: will apply

ui/src/hooks/api/useVmwareRevalidation.ts
sarika-pf9: Good catch on lines 76 and 104 — those can share a `useMemo`. However, line 50 (inside `refetchInterval`) must stay as-is: it reads from `query.state.data `directly to avoid react-query stale closure issues inside the interval callback. Using the `vmwareCreds` closure variable there would give outdated data. We'll apply useMemo for the useEffect and handleRefreshAndRevalidate lookups.

ui/src/hooks/api/useVmwareRevalidation.ts
sarika-pf9: The two states serve distinct purposes. Local `isRevalidating` controls the spinner immediately on click and owns the timeout — it's needed because `useMutation.isPending` goes false as soon as the POST returns 200, well before the backend finishes. The backend status drives continued polling if the component remounts mid-revalidation. Removing local state would break the immediate spinner feedback.

ui/src/hooks/api/useVmwareRevalidation.ts
sarika-pf9: The 500ms delay is intentional. The revalidation flow does two things server-side: (1) validates the VMware credential, then (2) fetches all VMware resources (VMs, clusters, datastores, networks) through that credential. The POST /revalidate_credentials returns 200 as soon as the process is kicked off — before the backend has written "Revalidating" to the CRD status. If we invalidateQueries immediately, the next poll returns the pre-revalidation "Succeeded" status with a fresh dataUpdatedAt timestamp, causing the hasFreshStatus && status !== 'Revalidating' completion check to fire instantly and incorrectly mark revalidation as done. The 500ms ensures the backend has had time to transition the CRD to "Revalidating" before the first poll. This is the same pattern in CredentialsTable.tsx (line 310–313).

AbhijeetThakur (approved):
LGTM

PR #1958, opened 2026-05-20
[1803] fix: Added retry for transient API errors

sarika-pf9:
## What this PR does / why we need it

Three classes of bugs addressed:

1. Missing cleanup after failure: ports created by ReservePortsForVM were orphaned when subsequent steps failed (CreateVolumes, StorageAcceleratedCopy path: InitializeStorageProvider, ValidatePrerequisites, CopyDisks, applyImageMetadata). Each failure path now calls cleanup() with portids and vcenterSettings so ports are deleted when CleanupPortsAfterMigrationFailure is enabled.

2. Transient API errors not retried in WaitForVolume: volumes.Get and servers.Get errors returned immediately, aborting the polling loop. The reported failure was a context deadline exceeded on servers.Get that exited all retries. Both calls now log and continue to the next iteration.

3. Nil OpenstackVol panic in cleanup: when CreateVolumes failed mid-loop, later disks had nil OpenstackVol. DetachAllVolumes and DeleteAllVolumes now skip nil entries so cleanup does not panic on partial volume creation.

Also: DeletePort and DeleteVolume now retry up to DeleteOperationRetryCount (5) times so transient errors during cleanup do not leave resources orphaned. Constant defined in pkg/common/constants alongside existing retry constants.

## Which issue(s) this PR fixes

fixes #1803 
fixes #1926

## Testing done
image

---

<a href="https://app.devin.ai/review/platform9/vjailbreak/pull/1958" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open in Devin Review
  </picture>
</a>

PR #2035, opened 2026-06-19
[1872] Verify volume bootable state after SetBootable timeout

### LINKED ISSUE #1872: Failed to set volume Bootable & Migration failed due to this

We tried to migrate VMs through vJailBreak v0.4.3 post conversion phase it mentioned as failed to convert disks:failed to make volume bootable
We faced this issue only in this version for 2 migrated VMs and out of that we could see one of the volumes are marked as ✅ bootable in UI.
After this migration had failed we could manually attach it and spin up a VM using same flow.

Image

Image

Migrate without issues due to make bootable.

### Environment

vJailBreak v0.4.3

### vCenter version

8.3

Maybe just a UI Bug/Fake call to vJB

sarika-pf9:
**Background — Issue #1872**
Users on vJailbreak v0.4.3 saw migrations fail with: `failed to set volume as bootable: Post "https://.../cinder/v3/.../action": context deadline exceeded (Client.Timeout exceeded while awaiting headers)`
The tell: one of the volumes showed as bootable in the OpenStack UI despite the migration being marked failed.

**What happened**
SetVolumeBootable made a single HTTP call to the Cinder API. If the response didn't arrive before the HTTP client timeout, it returned an error and failed the migration — even though Cinder had already committed the change server-side. The client simply never received the 200.

**In PR #1958** 
Added a retry loop (5 attempts, 5s sleep) around SetBootable. This handles transient Cinder slowness where the API is slow but eventually responds. Most intermittent timeout cases are resolved by retry 2 or 3.

Remaining edge case (this PR)
When all 5 retries time out, the old code still failed the migration. This can happen when Cinder is under sustained load for >25 seconds. Crucially, just like in the original bug, Cinder may have committed the change on any one of those attempts — the client just never got the response.

**Fix**
On each context deadline exceeded / Client.Timeout exceeded failure, immediately verify the actual volume state via GetVolume. If Cinder already set bootable=true, treat as success and return nil — no further retries needed. This is safe because SetBootable is idempotent.

fixes #1872 

---

<a href="https://app.devin.ai/review/platform9/vjailbreak/pull/2035" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


v2v-helper/pkg/utils/openstackopsutils.go
```diff
@@ -495,11 +507,27 @@ func (osclient *OpenStackClients) SetVolumeBootable(ctx context.Context, volume
...
+		// the HTTP response never reached us. Verify immediately so we don't
+		// retry an operation that already succeeded.
+		if isHTTPTimeoutError(err) {
+			current, getErr := osclient.GetVolume(ctx, volume.ID)
+			if getErr == nil && current.Bootable == "true" {
+				PrintLog(fmt.Sprintf("Volume %s is already bootable (Cinder applied the change but HTTP response timed out); treating as success", volume.ID))
+				return nil
+			}
+		}
+
```
spai-p9: isHTTPTimeoutError is a bit restrictive we can add this check irrespective of the error we get from the setbootable function. isHTTPTimeoutError is not needed, just check on err.

spai-p9 (changes requested):

spai-p9 (approved):
lgtm

PR #2049, opened 2026-06-25
[2050] Optimize retry flow and add bulk retry

### LINKED ISSUE #2050: Optimize retry flow, add bulk retry, and clean up migrations table UI
The retry feature had several reliability issues:

- Repeated retries of the same VM caused 409 conflicts (old -r suffixed template/plan name already existed)
- Retrying multiple VMs from the same plan caused 409 (all VMs shared the same template, so second retry tried to create the same -r template again)
- Migration DELETE returned 404 when Kubernetes GC cascade deleted the migration before the UI's DELETE call arrived
- If multiple VMs need to be retried without any configuration changes, users should have a bulk option to retry them all at once instead of triggering retries one by one.

sarika-pf9:
- Retry optimization: New plan and template are always created with fresh UUIDs (eliminating -r suffix conflicts on repeated retries). Old template is also deleted when the plan is the last VM, and migration DELETE is 404-tolerant to handle Kubernetes GC cascade.
- Bulk retry: Select multiple failed migrations in the table and retry all at once without editing configurations. A confirmation dialog shows before executing.
- Progress bar removed: Linear progress bar removed from the migrations table rows (detail page unchanged).

fixes #2050 

## Testing done

image
image


spai-p9 (approved):
lgtm.

PR #2061, opened 2026-06-29
docs: update hot-add proxy docs and vCenter permission requirements

meghansh-pf9:
Update hot-add proxy setup docs, fix snapshot name in troubleshooting, add Hot-Add as storage copy method, and add feature-specific vCenter permission tables.


spai-p9 (approved):
looks good.

PR #2108, opened 2026-07-07
Vcenter version limit for ova deployment

meghansh-pf9:
Document that the bundled Proxy VM OVA (vmx-21) requires ESXi/vCenter 8.0 U2 or newer for deployment.


sarika-pf9 (approved):
lgtm

PR #2110, opened 2026-07-07
[2085] Block Persist source network interfaces when mapped network subnet doesn't match VM IPs

### LINKED ISSUE #2085: Task: Block persist IP option if different network is selected
persist IP option does not make sense if user has selected different Network/Subnet. We should somehow block in UI. These two options should be mutually exclusive.

sarika-pf9:
The "Persist source network interfaces" option pins the source VM's MAC→IP mapping and NIC names inside the guest after migration. This makes no sense when the user maps a VMware network to a PCD network whose subnet does not contain the VM's IPs — the persisted IPs would be invalid on the target and the VM would lose connectivity. These two options are now mutually exclusive in the UI.

fixes #2085 

## Testing done

https://github.com/user-attachments/assets/0328474d-de74-4eae-9430-7e446862a915
image
image

### DIFF
--- a/ui/e2e/migration/edge-cases.spec.ts
+++ b/ui/e2e/migration/edge-cases.spec.ts
@@ -34,6 +34,8 @@ import {
   MOCK_RDM_DISKS_LIST,
   MOCK_SETTINGS_CONFIGMAP_NETWORK_PERSISTENCE_ON,
   MOCK_SETTINGS_CONFIGMAP_NETWORK_PERSISTENCE_OFF,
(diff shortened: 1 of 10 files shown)


OmkarDeshpande7 (commented):
@sarika-pf9 The better UX can be, block all the networks that do not have any subnet that can contain given VM IP. We have to think of multiple networks also. block until Fallback to DHCP is selected

OmkarDeshpande7 (approved):
lgtm

PR #2112, opened 2026-07-08
[2095] Serve debug bundle as .tar.gz archive instead of single .txt file

### LINKED ISSUE #2095: Enhancement: serve debug bundle as .tar.gz archive instead of single .txt file
The debug-bundle API (`GET /vpw/v1/debug-bundle`) returns one large plain-text `.txt` file with all sections concatenated: pod logs, every related resource YAML (separated by `FILE:` headers), and all `/var/log/pf9` debug logs. Real bundles reach tens of MB, which required raising the internal gRPC message limit and capping section sizes.

### Proposed enhancement

Return a compressed `.tar.gz` archive with a real directory structure:

sarika-pf9:
Changes the debug-bundle API response from one large concatenated `.txt` file to a compressed `.tar.gz` archive with a real directory structure.

fixes #2095 

## Testing done
image
image
image


meghansh-pf9 (approved):
lgtm

PR #2127, opened 2026-07-10
[1801] Twice HOTPLUG_CPU_MAX/HOTPLUG_MEMORY_MAX metadata so migrated VMs can be resized on PCD

### LINKED ISSUE #1801: Ability to set max vcpu and memory for Hot plug vms

As a user I should be able to provide the option to set max vcpu and max memory.
user may not want to set max and min as same always

### Detailed Description

In vjailbreak, hot plug migration sets the current values and the max value to be the same. This is unexpected. The only option for the user in this case is to recreate the vm with disks.

_No response_

sarika-pf9:
Removes the `HOTPLUG_CPU_MAX` and `HOTPLUG_MEMORY_MAX` server metadata keys that vJailbreak was setting during VM creation for hotplug (flavorless) migrations.

fixes #1801


OmkarDeshpande7 (approved):
lgtm, Can you please also change this in the docs?

PR #2133, opened 2026-07-13
[2012] Fix SUSE GRUB Legacy device references post-migration

### LINKED ISSUE #2012: Suse 11 sp4 boot failure post migration (pointing to old devices sda instead of vda)

The disk changed from sda to vda (VirtIO) but the configuration files still reference the old disk name. It stuck in ramfs shell.

Image

Image

Image

/dev/sda1    swap    swap    defaults    0 0   ← WRONG! should be vda
UUID=f808... /       ext3    acl,...     1 1   ← OK (uses UUID)

VM should boot post migration

### Environment

main

### vCenter version

NA

_No response_

meghansh-pf9:
fixed: #2012

Rewrite SUSE GRUB Legacy `root=`/`resume=` cmdline device references to UUID= without touching `device.map`, so migrated SUSE 11 SP4 guests boot instead of dropping to a ramfs shell.

__Testing done__
### VM booting successfully post migration

image

### DIFF
--- a/scripts/generate-mount-persistence.sh
+++ b/scripts/generate-mount-persistence.sh
@@ -94,19 +94,8 @@ get_fstab_id() {
   fi
 }
 
-# detect_bootloader <root_prefix>
-#
-# Prints "legacy", "grub2", or "unknown" for the guest rooted at
-# <root_prefix> (pass "" when not running inside a guestfish appliance, or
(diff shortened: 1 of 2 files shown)


spai-p9 (approved):
lgtm.

PR #2160, opened 2026-07-17
[2115] Fix IP address not populating on port creation due to MAC address case mismatch

### LINKED ISSUE #2115: IP address is not populating

Started migration on vjailbreak 0.4.6 and mapped the network resources correctly.

The migration attempts to continue, but the log shows that the IP address is not being populated (passing an empty array [] instead)

`2026/07/08 06:09:07 Port with MAC address 00:50:56:9D:47:74 does not exist, creating new port, trying with same IP address: []`

However, the IP address is visible and detected inside the vmwaremachine resource

vjailbreak should successfully extract the detected IP address (146.122.44.210) from the vmwaremachine resource and populate it when creating the new port, rather than leaving the IP array empty ([]).

### Environment

vjailbreak 0.4.6

### vCenter version

8.0.3

_No response_

sarika-pf9:
## Issue
During migration, the v2v-helper created Neutron ports with an empty IP list even though the IP was correctly detected and stored in the VMwareMachine resource. The port then fell back to DHCP, and on subnets with no free addresses the migration failed with `IpAddressGenerationFailure`.

## RCA

MAC address case mismatch between two sources:

- The **VMwareMachine CR** stores MACs in lowercase.
- **vCenter hardware config** preserves the case of manually-assigned MACs, so `vminfo.Mac` can contain uppercase (e.g. `00:50:56:9D:47:74`).

- `GetVMInfo` matched the two case-insensitively (`strings.EqualFold`) but keyed the `IPperMac` map with the CR's lowercase MAC, while port creation looked the map up with the vCenter uppercase MAC — the lookup missed and returned an empty IP list. The `NetworkInterfaces` fallback branch was worse: it compared MACs with case-sensitive `==` and would not match at all.

## Fix

Normalize MACs to canonical lowercase at every boundary

fixes #2115 

## Testing done

caps in MAC:
image
image

before:
image

after:
image

### DIFF
--- a/v2v-helper/pkg/utils/openstackopsutils.go
+++ b/v2v-helper/pkg/utils/openstackopsutils.go
@@ -636,7 +636,8 @@ func (osclient *OpenStackClients) CheckIfPortExists(ctx context.Context, ipEntri
 		return nil, err
 	}
 	for _, port := range portList {
-		if port.MACAddress == mac {
+		// Neutron stores MACs lowercase; compare case-insensitively
+		if strings.EqualFold(port.MACAddress, mac) {
 			if port.DeviceID != "" {
 				return nil, fmt.Errorf("precheck failed: port %s (MAC %s) is already in use by device %s", port.ID, mac, port.DeviceID)
 			}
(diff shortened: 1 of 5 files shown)


meghansh-pf9 (approved):
lgtm

OmkarDeshpande7 (approved):
lgtm

PR #2166, opened 2026-07-20
[2167] fix(v2v-helper): Install vendored libnbd with libnbd-devel

### LINKED ISSUE #2167: v2v-helper Docker build fails with dnf dependency resolution error
Image

sarika-pf9:
Copy and install the vendored `libnbd-1.25.5-1.fc44.x86_64.rpm` (already present in `v2v-helper/nbdkit/`) alongside `libnbd-devel` in the builder stage, so the exact-version dependency is satisfied from the command line regardless of repo state — same pattern the runtime stage uses.

fixes #2167 

## Testing done
image


meghansh-pf9 (approved):
LGTM

PR #2168, opened 2026-07-20
Refactoring boot partition detection 

meghansh-pf9:

Adding safety checks to avoid runtime panics over out of bound values and improving logging mechanism for boot partition detection script"
command inside the guestfish appliance.


OmkarDeshpande7 (approved):
lgtm

PR #2190, opened 2026-07-23
[2179] fix(migration): recompute subnet-mismatch warning when IP changes after network mapping

### LINKED ISSUE #2179: Error message not getting updated if user updated the IP after selecting the network

Error message not getting updated if user updated the IP after selecting the network -

https://github.com/user-attachments/assets/910a53d8-97d2-4bd3-b4bc-5b18ae40eaee

Error message should be updated if user updated the IP after selecting the network

### Environment

main

### vCenter version

NA

_No response_

sarika-pf9:
**RCA:** useNetworkSubnetCompatibility.ts gated its recompute effect on mappingsKey — a signature built only from network mapping source|target pairs. Editing or clearing a VM's IP after the mapping was already made doesn't change that key, so the effect's if (mappingsKey === prevMappingsRef.current) return guard short-circuits and subnetWarnings never recomputes. The banner (and the "Persist source network interfaces" gate that reads the same state) keeps showing the stale IP/message indefinitely, even after the IP is updated to something now compatible (or incompatible) with the mapped subnet.

**Fix:** Extracted computeSubnetCheckSignature(mappings, networkIPsMap) — folds each mapping's resolved IPs (sorted, for order-independence) into the dedup key alongside the mapping pairs. Any IP edit/clear now changes the signature, so the effect re-runs and rebuilds subnetWarnings from scratch, correctly dropping or updating the affected network's message. checkNetworkSubnetCompatibility API-response caching (apiCacheRef, keyed by target+IPs) is untouched — this only fixes the outer gate that was skipping recompute entirely.

fixes #2179

## Testing done

https://github.com/user-attachments/assets/c5291944-8431-43cc-b66b-e432ad46aaef

### DIFF
--- a/ui/src/features/migration/hooks/useBulkIPEdit.test.ts
+++ b/ui/src/features/migration/hooks/useBulkIPEdit.test.ts
@@ -1,0 +1,106 @@
+import type { SetStateAction } from 'react'
+import { describe, expect, it, vi } from 'vitest'
+import { renderHook, act } from '@testing-library/react'
+import { buildClearAllIpsUpdate, useBulkIPEdit } from './useBulkIPEdit'
+import type { VmDataWithFlavor } from '../types'
+
+describe('buildClearAllIpsUpdate', () => {
+  it('empties the IP field and turns Preserve IP off for every interface', () => {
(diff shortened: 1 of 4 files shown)


AbhijeetThakur (approved):
LGTM

PR #2192, opened 2026-07-23
[2213] Fix migration stuck in ConvertingDisk after transient VM creation timeout

### LINKED ISSUE #2132: migration CR remains in "ConvertingDisk" state after successful VM migration following transient failure

A VM migration encountered a transient failure to get the VM status from PCD, but eventually completed successfully. The migration job and pod completed successfully, and the VM was brought up and running in PCD. However, the Migration CR was not updated after the successful completion and remained stuck in the Converting phase indefinitely.

Once the migration completed successfully, the Migration CR should have been updated to the Succeeded state instead of remaining in the Converting phase.

### Environment

Vjailbreak 0.4.7

### vCenter version

8.0.3

_No response_

OmkarDeshpande7:
## Summary

- `verifyVMCreatedDespiteTimeout` now returns `(string, error)` so caller gets the confirmed server ID
- Recovery path uses `utils.PrintLog` for error context (no K8s Event) and emits `logMessage("VM created successfully: ID: %s")` so the controller transitions correctly
- `CreateFailedCondition` now matches `EventMessageMigrationFailed` ("Trying to perform cleanup") instead of loose `"failed to"` substring — eliminates false Failed conditions on non-fatal warnings
- Non-fatal cleanup warnings in `hotadd_copy.go`, `vaai_copy.go`, and post-success `DisconnectSourceNetworkIfRequested` downgraded from `logMessage` to `utils.PrintLog` to avoid spurious K8s Events

## Root Cause

When `CreateTargetInstance` timed out but the VM was confirmed ACTIVE via `verifyVMCreatedDespiteTimeout`, the recovery path called `logMessage` with `"failed to"` in the message. This:
1. Created a Kubernetes Event containing `"failed to"` → triggered the loose `CreateFailedCondition` check
2. Never emitted `"VM created successfully"` → controller never saw the success signal → migration stuck at `ConvertingDisk`

## Test plan

- [x] Unit tests added in `migrationutils_test.go` covering the fixed condition matching logic
- [x] Run `cd k8s/migration && make test`
- [x] Verify migration completes when `CreateTargetInstance` times out but VM becomes ACTIVE

Closes #2132

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/k8s/migration/pkg/utils/migrationutils.go
+++ b/k8s/migration/pkg/utils/migrationutils.go
@@ -128,7 +128,7 @@ func CreateMigratingCondition(migration *vjailbreakv1alpha1.Migration, eventList
 func CreateFailedCondition(migration *vjailbreakv1alpha1.Migration, eventList *corev1.EventList) []corev1.PodCondition {
 	existingConditions := migration.Status.Conditions
 	for i := 0; i < len(eventList.Items); i++ {
-		if eventList.Items[i].Reason != constants.MigrationReason || !strings.Contains(eventList.Items[i].Message, "failed to") {
+		if eventList.Items[i].Reason != constants.MigrationReason || !strings.Contains(eventList.Items[i].Message, constants.EventMessageMigrationFailed) {
 			continue
 		}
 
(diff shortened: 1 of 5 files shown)


spai-p9 (approved):
looks good.

PR #2205, opened 2026-07-24
[2201] Show migration failure messages on the UI

### LINKED ISSUE #2201: Error msg is not visible either on migrations page or on details page
Image

Image

sarika-pf9:
**RCA:** PR #2192 : Narrowed the match to only "Trying to perform cleanup". Side effect: broke visibility for almost every other failure type.

**Fix:** This PR restores broad detection of genuine failures while still correctly ignoring the one harmless recovery case the earlier fix was meant to handle, and also filters out non-fatal warning messages so they don't get mistaken for failures. As a result, users again see the actual error message for a failed migration, without reintroducing the original stuck-migration problem.

fixes #2201 

## Testing done

https://github.com/user-attachments/assets/03dd11bd-0370-437a-960c-05c66883b69a

### DIFF
--- a/k8s/migration/pkg/utils/migrationutils.go
+++ b/k8s/migration/pkg/utils/migrationutils.go
@@ -123,19 +123,12 @@ func CreateMigratingCondition(migration *vjailbreakv1alpha1.Migration, eventList
 	return existingConditions
 }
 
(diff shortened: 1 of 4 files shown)


OmkarDeshpande7 (approved):
lgtm

PR #2206, opened 2026-07-24
[docs] Network traffic separation guide for dedicated data network (VJAILB-85)

### LINKED ISSUE #2207: Network documentation update for vjailbreak
## Issue

In some customer deployments, the VMware underlay architecture — including the storage network — differs from the production PCD host underlay, often by design to meet security requirements. Planning and segregating network traffic (e.g., the storage block-copy path, multi-NIC configuration) in vjailbreak isn't clearly documented today.

## Expected result

Clear documentation on how network traffic can be effectively segregated (management vs. storage) within vjailbreak, for customer and deployment reference. Should include a flow diagram and recommendations customers can follow to set this up.

## Design notes (traffic separation)

The core idea is keeping vCenter's control-plane traffic and vjailbreak's data-copy traffic on physically distinct paths. On the ESXi side, this is enforced at the interface level — one interface is tagged for NFC (Network File Copy, used for disk data streaming during copy/migration), the other handles mgmt/API traffic. On the vjailbreak side, this is mirrored: one bridge/bond carries vCenter management traffic, the other is dedicated to the storage network used for actual disk data copy.

The critical detail is routing on the vjailbreak VM itself: when it issues a data-copy request, ESXi responds via the NFC-tagged interface's IP. If vjailbreak's outbound route for that request isn't explicitly pinned to the storage-network interface, the request can go out via the mgmt interface instead — causing an IP mismatch with ESXi's NFC response path, leading to asymmetric routing, dropped responses, or data-copy traffic silently falling back onto (and congesting) the management network. The routing table must explicitly route NFC/data-copy traffic through the storage-network interface, not the default route.

OmkarDeshpande7:
## Summary

- Adds new How-to guide: **Configuring Dedicated Data Network for Migrations**
- Documents the ESXi iface1 (mgmt) / iface2 (NFC data copy) split
- Documents vJailbreak VM Bridge-1/Bond0 (mgmt) vs Bridge-2/Bond1 (storage) design
- Covers the critical routing table requirement to prevent asymmetric routing when ESXi responds via iface2 IP
- Includes the architecture diagram from VJAILB-85

## Test plan

- [ ] Verify image renders correctly in the docs site
- [ ] Verify page appears under Guide → How-to in sidebar (autogenerated)
- [ ] Review technical accuracy of routing section against actual deployment

Closes #2207

🤖 Generated with [Claude Code](https://claude.com/claude-code)


docs/src/content/docs/guides/How-to/network-traffic-separation.md
```diff
@@ -0,0 +1,81 @@
...
+
+When vJailbreak issues a disk-copy request, ESXi responds using **iface2's IP** — because iface2 is the interface tagged for NFC traffic. If the vJailbreak VM's outbound route for that request is not explicitly pinned to the storage interface (Bridge-2 / Bond1), the following problems occur:
+
+- **Asymmetric routing** — request exits via the management interface, response arrives via the storage interface; the connection is dropped.
+- **Silent fallback** — data copy traffic falls back onto the management network, starving it of bandwidth.
+- **Transfer failures** — the NBD copy fails or times out due to IP mismatch on ESXi's NFC response path.
+
+### Required Routing Rule
+
+On the vJailbreak VM, add an explicit route so that any traffic destined for ESXi's NFC/data-copy service (iface2's IP subnet) is routed out through the storage interface:
```
sarvindpf9, line 48: imho, this should be optional because when the second interface with an IP is added to the vjb VM that is in the range of NFC applied subnet, this interface will have a default entry which essentially will be used when trying to reach the hypervisor. Unless any additional L3 routing is required to reach the hypervisor IP even being placed in the same subnet, adding additional routes would not be required.

OmkarDeshpande7: Agreed. IMO this won't harm to do add it always. There can be multiple cases where absence of this route can cause asymmetric routing which will lead to migration failures (example, above comment makes an assumption that NFC network is a different subnet).

spai-p9 (approved):
lgtm.

PR #2208, opened 2026-07-24
Support encrypted volume types in pcd. 

### LINKED ISSUE #1622: vJailbreak need to work with VMs encrypted with Akeyless

When VMs on VMWare are encrypted with Akeyless(https://www.akeyless.io/), vJailbreak by default won't be able to ready/get any information about the VMs like what OS it has, IP, and won't be having access to the .vmdk files.
VMs require access in order to boot, and thus, decryption must occur PRIOR to any authentication/authorization to the virtual machines themselves.

### Detailed Description

As explained above.

_No response_

spai-p9:
vjailbreak's copy step trusted the destination was pre-zeroed (nbdcopy --target-is-zero, plus hole-punching on incremental syncs), so it skipped writing zero/sparse regions. That's fine for plain volumes, but on encrypted volumes those skipped regions were never written through the QEMU encryption layer — so they decrypt to garbage instead of zero. Real data copied fine; sparse regions like the partition table came back corrupted.

Fix: Detect when the destination volume is encrypted and do a full dense copy — no --target-is-zero, and write real zeros instead of punching holes on syncs — so every region passes through the encryption engine correctly. Unencrypted migrations are unchanged.

fixes #1622 

## Testing done

### DIFF
--- a/v2v-helper/nbd/nbdops.go
+++ b/v2v-helper/nbd/nbdops.go
@@ -32,7 +32,7 @@ type NBDOperations interface {
 	StartNBDServer(vm *object.VirtualMachine, server, username, password, thumbprint, snapref, file string, progchan chan string) error
 	StopNBDServer() error
 	CopyDisk(ctx context.Context, dest string, diskindex int, destEncrypted bool) error
-	CopyChangedBlocks(ctx context.Context, changedAreas types.DiskChangeInfo, path string) error
+	CopyChangedBlocks(ctx context.Context, changedAreas types.DiskChangeInfo, path string, destEncrypted bool) error
 	GetProgress() (int64, int64, time.Duration)
 }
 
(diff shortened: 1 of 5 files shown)


v2v-helper/nbd/nbdops.go
```diff
@@ -508,18 +525,49 @@ func zeroRange(fd *os.File, offset int64, length int64) error {
...
+	flags := uint32(unix.FALLOC_FL_PUNCH_HOLE | unix.FALLOC_FL_KEEP_SIZE)
+	punchErr := syscall.Fallocate(int(fd.Fd()), flags, offset, length)
+	if punchErr == nil {
+		return nil
+	}
+	// Fall back to writing zeros directly. This is the slow path; it
+	// triggers on filesystems that don't support FALLOC_FL_PUNCH_HOLE
+	// (some NFS configurations, tmpfs, certain block-backed targets).
+	utils.PrintLog(fmt.Sprintf("Failed to punch hole at offset %d, falling back to pwrite: %v", offset, punchErr))
+	utils.PrintLog(fmt.Sprintf("Unable to zero range %d - %d on destination, falling back to pwrite: %v", offset, offset+length, punchErr))
```
OmkarDeshpande7, line 561: please remove the print statements, it will make the process slow as this function is called in loop

spai-p9: This printLog already existed.

OmkarDeshpande7 (approved):
lgtm

PR #2224, opened 2026-07-28
[2212] Refactoring migration scheduling and validations

### LINKED ISSUE #2212: Scheduling Issue in Agents

- Migrations only being scheduled on agent nodes despite master node having compute capacity
- Scaling after triggering migrations(stuck in pending state) does not result in them being scheduled on the newly created agent
- Phase not being updated for all migrations upon creating a migrationplan with large number of VMs together, resulting in migrations showing up in unknown state on UI

Image

#### All agents are available but still migrations are stuck in pending/Uknown state

Image

- Proper Status update for each migration
- Migrations shall be scheduled on idle nodes having appropriate capacity, instead of staying in idle state

### Environment

NA

### vCenter version

NA

NA

meghansh-pf9:
A VM with no matching OpenStack flavor used to abort the whole MigrationPlan — every VM behind it in the list never got a ConfigMap or Job, and the plan retried forever on the same VM.

Now the failure is contained to that VM:
- Flavors are resolved in pre-flight validation, once per plan instead of once per VM (and lazily, so operator-pinned plans still never call Nova).
- `ErrNoSuitableFlavor` sentinel + `errors.Is` so only an unschedulable shape is treated as a skip; auth/connectivity errors still requeue. Replaces a `strings.Contains` check.
- The unschedulable VM gets a `Migration` in `ValidationFailed` with an actionable reason, so it stays visible in the UI (no UI changes needed).
- `processMigrationPhases` accumulates per-VM failures instead of returning on the first one; the plan is marked `PodFailed` only once nothing is still running, so post-migration still runs for the VMs that succeeded.

fixes #2212

__Testing Done__

image

### DIFF
--- a/k8s/migration/internal/controller/migrationplan_controller.go
+++ b/k8s/migration/internal/controller/migrationplan_controller.go
@@ -28,6 +28,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	"github.com/gophercloud/gophercloud/v2/openstack/compute/v2/flavors"
 	"github.com/pkg/errors"
 	vjailbreakv1alpha1 "github.com/platform9/vjailbreak/k8s/migration/api/v1alpha1"
 	"github.com/platform9/vjailbreak/k8s/migration/pkg/scope"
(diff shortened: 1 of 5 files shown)


k8s/migration/internal/controller/migrationplan_controller.go
```diff
@@ -71,6 +72,10 @@ const VDDKDirectory = "/home/ubuntu/vmware-vix-disklib-distrib"
 // StorageCopyMethod is the storage copy method value for Storage Accelerated copy
 const StorageCopyMethod = "StorageAcceleratedCopy"
 
+// targetFlavorIDKey is the migration ConfigMap key holding the OpenStack flavor ID
+// the VM will be created with. Read back by v2v-helper as TARGET_FLAVOR_ID.
+const targetFlavorIDKey = "TARGET_FLAVOR_ID"
```
OmkarDeshpande7: move this to common constants module

OmkarDeshpande7:
`k8s/migration/internal/controller/migrationplan_controller.go` — `planNeedsFlavorLookup` calls `GetVMwareMachineForVM` once per VM just to check `TargetFlavorID`, and `validateMigrationPlanVMs` immediately re-fetches the same objects again. Doubles Get calls per VM before any Nova lookup (mitigated somewhat by client caching, but still redundant).

Fix: have `planNeedsFlavorLookup` return the fetched machines for reuse instead of a bare bool.

OmkarDeshpande7:
`k8s/migration/internal/controller/migrationplan_controller.go` — `if validation != nil { ... }` guards a value that can never be nil at that point (initialized to a non-nil struct before use; `validateMigrationPlanVMs` only returns nil alongside a non-nil error, which triggers an earlier return). Per project guidance, avoid error handling for scenarios that can't happen.

Fix: drop the nil check.

pkg/common/openstack/flavor.go
```diff
@@ -11,6 +12,17 @@ import (
...
+// ErrNoSuitableFlavor indicates that no flavor in the candidate set can satisfy a
+// VM's CPU / memory / GPU shape. This is a property of that one VM, not of the
+// cluster, so callers may treat it as a per-VM skip rather than a fatal error —
+// match it with errors.Is instead of string-matching the message.
+//
+// Note that an empty candidate set is deliberately NOT reported as this sentinel:
+// that means Nova returned no flavors at all (a credentials, connectivity or
+// availability-zone-filter problem) and must stay fatal, because silently
+// skipping it would skip every VM in the plan.
+var ErrNoSuitableFlavor = errors.New("no suitable flavor found")
```
OmkarDeshpande7, line 24: move to const

OmkarDeshpande7 (approved):
lgtm

PR #2225, opened 2026-07-29
[2195] Fix incorrect  step timings on the migration detail page

### LINKED ISSUE #2195: Converting disk time is incorrect on detail page

Converting disk time is incorrect on detail page -

Image

Converting disk time should be correct

### Environment

main

### vCenter version

NA

_No response_

sarika-pf9:

- Each stepper step now shows its own duration (delta between boundary conditions), not cumulative time since migration creation.
- Converting Disk no longer reuses the same Migrating condition as Cutover (was producing identical, wrong durations for both) — now bounded by Migrated/DataCopied.
- Multi-disk DataCopy boundary now takes the max lastTransitionTime across all per-disk entries instead of an arbitrary first-array-match, which could silently pick an earlier disk's timestamp.
- Backend: Pending→Validating phase transition now gates on the pod actually being Running/terminal, so Pending reflects real agent-queue wait time instead of bleeding into Validating.
- Backend: added PodRunning condition (from the pod's native Status.StartTime) so Pending/Validating durations split correctly, including after the migration finishes.
- Backend: added CutoverTriggered condition (matches "Admin cutover triggered" events) so Cutover shows real trigger→conversion-start time for admin-gated migrations instead of the entire admin-wait period.
- Active/paused/failed step elapsed now uses each step's own start boundary instead of always cumulative-since-creation; paused (awaiting admin, not yet triggered) intentionally kept showing total-since-creation.
- Status badge no longer shows "Unknown" for a freshly-created migration before status populates — defaults to "Pending" (backend never actually assigns Phase Unknown).
- "Migration Complete" success banner duration formatting fixed — was dropping seconds under an hour and forcing all-caps h/m/s; now reuses the shared formatter with correct casing.
- Detail page subtitle now shows the real vSphere datacenter instead of the VMwareCreds object's name (e.g. "vmware").
- Events tab timeline now includes Created, Migration Started, and Cutover entries (previously only Validated/DataCopy/Migrating/Migrated).
- Added unit test coverage for all of the above.

fixes #2195

## Testing done
image
image
image
image
image
image
image
image

### DIFF
--- a/k8s/migration/internal/controller/migration_controller.go
+++ b/k8s/migration/internal/controller/migration_controller.go
@@ -194,13 +194,14 @@ func (r *MigrationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		}
 	}
 
-	if constants.VMMigrationStatesEnum[migration.Status.Phase] <= constants.VMMigrationStatesEnum[vjailbreakv1alpha1.VMMigrationPhaseValidating] {
-		migration.Status.Phase = vjailbreakv1alpha1.VMMigrationPhaseValidating
-	}
-	// Check if the pod is in a valid state only then continue
-	if pod.Status.Phase != corev1.PodRunning && pod.Status.Phase != corev1.PodFailed && pod.Status.Phase != corev1.PodSucceeded {
+	// Stay Pending until the pod has actually started running (or reached a terminal
+	// state).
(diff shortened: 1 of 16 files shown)


OmkarDeshpande7 (approved):
lgtm

AbhijeetThakur (approved):
UI changes LGTM

PR #2232, opened 2026-07-30
[2257] Fix migration failure when the guest root filesystem spans multiple disks

### LINKED ISSUE #2257: BTRFS boot partitions spanning over multiple disks failing

Created a suse VM with root partition spanning over multiple disks failing in the conversion process with multi boot device not supported

Correct bootable disk shall be detected

### Environment

main

### vCenter version

NA

_No response_

meghansh-pf9:

Guests whose root filesystem spans several devices fail with:

    guestfish: multi-boot operating systems are not supported

Reproduced on openSUSE Leap 15.5 after `btrfs device add /dev/sdb /`.

Every guest operation used `guestfish -i`, which aborts whenever `inspect_os()`
returns more than one root. It counts roots, it does not compare them — and
every btrfs member device carries a full superblock, with no `btrfs_member`
type for libguestfs' `_member` filter to catch. So each member is reported as a
separate root.

Resolve mounts explicitly instead of using `-i`: enumerate roots without `-i`,
collapse roots sharing a filesystem UUID (for btrfs, the `fsid`), then mount
shortest-path-first with a fatal `/` and best-effort everything else, matching
`inspect_mount_root`. Genuine multi-boot guests still fail, now naming the roots.

Also batches independent guestfish commands into single appliance boots, and
drops the per-disk boot probe, which cannot work when the root spans disks and
whose result was discarded anyway.

Fixes: #2257

### DIFF
--- a/v2v-helper/migrate/migrate.go
+++ b/v2v-helper/migrate/migrate.go
@@ -1174,18 +1174,6 @@ func (migobj *Migrate) LiveReplicateDisks(ctx context.Context, vminfo vm.VMInfo)
 	return vminfo, nil
 }
 
-// getBootCommand returns the appropriate command to detect boot volume based on OS type
-func (migobj *Migrate) getBootCommand(osType string) string {
-	switch strings.ToLower(osType) {
-	case constants.OSFamilyWindows:
-		return "ls /Windows"
-	case constants.OSFamilyLinux:
-		return "ls /boot"
-	default:
-		return "inspect-os"
-	}
-}
-
 // attachAllVolumes attaches all volumes and updates their paths in vminfo
 func (migobj *Migrate) attachAllVolumes(ctx context.Context, vminfo *vm.VMInfo) error {
 	for idx, vmdisk := range vminfo.VMDisks {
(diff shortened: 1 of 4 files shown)


v2v-helper/migrate/migrate.go
```diff
@@ -1198,28 +1198,40 @@ func (migobj *Migrate) attachAllVolumes(ctx context.Context, vminfo *vm.VMInfo)
...
-	for idx := range vminfo.VMDisks {
-		ans, cmdErr := virtv2v.RunCommandInGuest(vminfo.VMDisks[idx].Path, getBootCommand, false)
-		if cmdErr != nil || ans == "" {
-			continue
-		}
-
-		utils.PrintLog(fmt.Sprintf("Boot volume detected: Disk %d (%s)", idx, vminfo.VMDisks[idx].Name))
-		osPath = strings.TrimSpace(ans)
-		bootVolumeIndex = idx
-		break
```
OmkarDeshpande7, line 1216: @meghansh-pf9 let's not change what is working today. It is okay to have this redundant check, this is what is causing LDM data disk migrations to go through successfully today

OmkarDeshpande7:
`v2v-helper/virtv2v/virtv2vops.go` — `buildArgs` in `RunNetworkPersistence` constructs guestmount `-m` arguments by joining device+mountpoint with `:`:

```go
mountArg := spec.Device + ":" + spec.MountPoint
```

`guestmount -m` expects `dev[:mountpoint[:options[:fstype]]]` — it splits on the **first** colon. When `inspect-get-mountpoints` returns btrfs subvolume mountables (e.g., `btrfsvol:/dev/sda6/@/home`), the resulting arg `btrfsvol:/dev/sda6/@/home:/home` is mis-parsed: guestmount sees device=`btrfsvol`, mountpoint=`/dev/sda6/@/home`.

For openSUSE Tumbleweed with btrfs subvolumes — the guest class this PR targets — `inspect-get-mountpoints` typically returns btrfsvol mountables for `/` and nested mounts. The first attempt fails, triggering the fallback (root-only). Fallback is generally safe (network scripts write to `/etc`) but nested filesystems won't be mounted, which can silently mis-configure guests with split-subvolume layouts.

Fix: Limit guestmount `-m` args to plain-device mounts only (exclude btrfsvol mountables) — the fallback already does root-only which is sufficient for network persistence. Or consult guestmount(1) for the correct way to pass btrfsvol device paths.
---
`v2v-helper/migrate/migrate_test.go` — `detectBootVolume` changed substantially: it now probes all disks together via `RunCommandInGuestAllVolumes` instead of probing one disk at a time, and always returns `bootVolumeIndex = -1` (the OS-specific handlers re-derive the index from scratch). The behavioral contract changed but there are no new tests for this function.

Project constitution §IV requires tests for every modified Go file. `detectBootVolume` calls `virtv2v.RunCommandInGuestAllVolumes` which execs guestfish — to make it testable, inject the guestfish caller via a function variable or small interface, then write table-driven tests verifying: all-volumes probe, correct `osPath` returned, -1 index returned for both success and error paths.

OmkarDeshpande7:
`v2v-helper/virtv2v/virtv2vops.go` — `buildArgs` in `RunNetworkPersistence` constructs guestmount `-m` arguments by joining device+mountpoint with `:`:

```go
mountArg := spec.Device + ":" + spec.MountPoint
```

`guestmount -m` expects `dev[:mountpoint[:options[:fstype]]]` — it splits on the **first** colon. When `inspect-get-mountpoints` returns btrfs subvolume mountables (e.g., `btrfsvol:/dev/sda6/@/home`), the resulting arg `btrfsvol:/dev/sda6/@/home:/home` is mis-parsed: guestmount sees device=`btrfsvol`, mountpoint=`/dev/sda6/@/home`.

For openSUSE Tumbleweed with btrfs subvolumes — the guest class this PR targets — `inspect-get-mountpoints` typically returns btrfsvol mountables for `/` and nested mounts. The first attempt fails, triggering the fallback (root-only). Fallback is generally safe (network scripts write to `/etc`) but nested filesystems won't be mounted, which can silently mis-configure guests with split-subvolume layouts.

Fix: Limit guestmount `-m` args to plain-device mounts only (exclude btrfsvol mountables) — the fallback already does root-only which is sufficient for network persistence. Or consult guestmount(1) for the correct way to pass btrfsvol device paths.

OmkarDeshpande7:
`v2v-helper/migrate/migrate_test.go` — `detectBootVolume` changed substantially: it now probes all disks together via `RunCommandInGuestAllVolumes` instead of probing one disk at a time, and always returns `bootVolumeIndex = -1` (the OS-specific handlers re-derive the index from scratch). The behavioral contract changed but there are no new tests for this function.

Project constitution §IV requires tests for every modified Go file. `detectBootVolume` calls `virtv2v.RunCommandInGuestAllVolumes` which execs guestfish — to make it testable, inject the guestfish caller via a function variable or small interface, then write table-driven tests verifying: all-volumes probe, correct `osPath` returned, -1 index returned for both success and error paths.

OmkarDeshpande7:
`v2v-helper/virtv2v/guestfish.go:498` — `os.Setenv("LIBGUESTFS_BACKEND", "direct")` in `runGuestfishScript` is a process-wide, non-atomic write. If two migrations run concurrently, both goroutines can race on the env var. The same pre-existing pattern appears in `RunCommandInGuest` and `RunCommandInGuestAllVolumes`, but adding it to the new plan-resolution path widens exposure.

Prefer setting it once at process startup or using `cmd.Env = append(os.Environ(), "LIBGUESTFS_BACKEND=direct")` on the cmd struct directly, which scopes the variable to that one exec and is goroutine-safe.

PR #2233, opened 2026-07-30
[2234] Fix: remove VM selection from template creation and refactor pruning of mappings

### LINKED ISSUE #2234: Migration Template: remove VM selection from template creation and fix options not restoring on Use
Creating a Migration Template currently reuses the New Migration form as-is, and applying a
template doesn't restore everything that was saved.

1. **VM selection (step 2) is shown while creating a template**, along with its VM warnings —
   a template is VM-agnostic, so the step doesn't belong there.
2. **Mappings and image profiles are destroyed on "Use template".** They're pruned against the
   selected VMs' networks/datastores, but at the moment a template is applied no VMs are selected
   yet, so everything gets pruned away before the operator can pick VMs. Selecting a VM afterwards
   has nothing left to restore.
3. **Some options never come back.** Data copy method, cutover and periodic sync revert to
   defaults, and Windows profiles don't get selected even when a Windows VM is chosen.
4. **Fields that exist in the CRD are missing from the UI wiring**, so they're neither saved nor
   shown: `preserveSourceTags`, `customMetadata`, cutover window times (`vmCutoverStart` /
   `vmCutoverEnd`). Storage-accelerated templates also show no storage in the details panel,
   because datastore→ArrayCreds mappings aren't rendered or counted.

sarika-pf9:
- **Removed VM selection from template creation.** 

- **Fixed mappings and profiles being destroyed on "Use template".** 

- **Added the missing fields that existed in the CRD but were never wired into the UI**.

- **Fixed data copy method / cutover / periodic sync reverting to defaults**.

fixes #2234


OmkarDeshpande7 (commented):
PR #2233 review — [2234] Fix: remove VM selection from template creation and refactor pruning of mappings

Verdict: APPROVE

Well-structured fix for template UX bugs (mappings destroyed on "Use template", missing fields round-tripping, options reverting to defaults). The mapping-pool pattern is the right abstraction — keeping the full saved mapping set alive separately from the pruned form state, then re-applying lazily as sources appear. Test coverage is exceptional: the hook-level `driveHook` harness and `reconcileImageProfiles` settlement tests are exactly the right level for this class of UI state bugs.

Five minor findings below. None are blocking.

OmkarDeshpande7:
`ui/src/features/migration/steps/SecurityGroupAndServerGroup.tsx:118` — Guard ordering is non-obvious

The two early-return guards interact in a subtle way:

```ts
if (selectedImageProfiles.length === 0 && !templateImageProfiles?.length) return
// ...
if (!params?.vms || params.vms.length === 0) return
```

When a template is applied (`templateImageProfiles` is set, `selectedImageProfiles` is empty), the first guard is skipped and execution falls through to the `params.vms` guard. That's correct. But the natural reading is "no vms = nothing to do" — which should logically come first.

No bug, but consider swapping order for clarity:

```ts
// Check VM selection first — nothing to reconcile until VMs are chosen
if (!params?.vms || params.vms.length === 0) return
// Then check whether there's anything to prune or apply
if (selectedImageProfiles.length === 0 && !templateImageProfiles?.length) return
```

Fix: Swap the two guard conditions.

OmkarDeshpande7:
`ui/src/features/migration/hooks/useFilteredMappings.ts:170` — Spurious dep for HotAdd mode

`filteredStorageMappings` is in the dep array of the storage effect, but for HotAdd mode the code uses `params.storageMappings || []` as the base — the filtered value is only consumed in `'normal'` mode:

```ts
const base = storageCopyMethod === 'normal' ? filteredStorageMappings : params.storageMappings || []
```

Any change to `filteredStorageMappings` (e.g. a recompute when vmWareStorage changes) will trigger the effect for HotAdd unnecessarily. The `mappingsEqual` guard prevents incorrect writes, so this is purely cosmetic, but it could cause an extra render cycle.

Fix: Remove `filteredStorageMappings` from the dep array, or split the effect by storage mode.

OmkarDeshpande7:
`ui/src/features/migration/api/migration-blueprints/adapters.test.ts:127` — Personal name in test fixture

```ts
firstBootScript: 'echo "sarika"'
```

This is a personal name committed to a test. Harmless, but test fixtures should use generic sentinels.

Fix: `'echo "hello"'` or `'#!/bin/bash\necho done'`

OmkarDeshpande7:
`ui/src/features/migration/hooks/useFormValidation.ts` — Verify `VmData.networks/datastores` are typed as optional

`deriveSourceCatalog` uses `vm[key] || []` to handle missing fields:

```ts
return uniq(flatten(vms.map((vm) => vm[key] || []))).sort(stringsCompareFn)
```

The test at line 1578 exercises a `VmData` with `networks` absent:

```ts
const vms: VmData[] = [{ id: 'a', name: 'a', datastores: ['datastore1'] }]
```

This only compiles if `networks` is optional in `VmData`. Quick confirmation that `VmData.networks` and `VmData.datastores` are declared `string[]?` (not `string[]`) would ensure the defensive `|| []` is intentional and type-safe.

OmkarDeshpande7 (approved):
lgtm

PR #2258, opened 2026-08-05
[docs] Add "Retry a Failed Migration" guide for retry mechanism

### LINKED ISSUE #1750: Modifying Existing Migrations in Terms of Flavor, Datastore, or Other Settings

sarika-pf9:
Documents the retry and edit-and-retry workflow for failed migrations

fixes #1750

### DIFF
--- a/docs/src/content/docs/reference/known-limitations.md
+++ b/docs/src/content/docs/reference/known-limitations.md
@@ -241,3 +241,18 @@ Hot migration (**Copy live VMs, then power off**) minimizes downtime but still r
 :::caution
 Before migrating, verify that your application starts cleanly after a cold reboot. Applications that require manual intervention to restart (e.g., databases with crash-inconsistent state) should be cleanly shut down inside the VM before initiating cold migration.
 :::
+
+## Retrying a Failed Migration
+
+The **Retry** action reopens a failed migration in the migration form so its configuration can be corrected. The following limitations apply. See [Retry a Failed Migration](../../guides/how-to/retry_failed_migration/) for the full workflow.
+
+| Limitation | Detail |
+|---|---|
+| VMs with RDM disks cannot be retried | Shared RDM disk state prevents an automatic retry. The Retry button is disabled for these migrations and the migration must be restarted manually. |
+| A retry always produces a single-VM plan | Retrying one VM from a plan that covered several VMs moves that VM into a plan of its own. The remaining VMs stay in the original plan and are unaffected. |
+| Bulk retry cannot change configuration | **Retry Selected** restarts each migration with its existing configuration. To change settings, retry the migration individually. |
+| Credentials and source cluster are locked | A retry cannot change the VMware or OpenStack credentials or the source cluster. Create a new migration instead. |
+
+:::note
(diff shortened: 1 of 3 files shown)


docs/src/content/docs/reference/known-limitations.md
```diff
@@ -241,3 +241,18 @@ Hot migration (**Copy live VMs, then power off**) minimizes downtime but still r
...
+## Retrying a Failed Migration
+
+The **Retry** action reopens a failed migration in the migration form so its configuration can be corrected. The following limitations apply. See [Retry a Failed Migration](../../guides/how-to/retry_failed_migration/) for the full workflow.
+
+| Limitation | Detail |
+|---|---|
+| VMs with RDM disks cannot be retried | Shared RDM disk state prevents an automatic retry. The Retry button is disabled for these migrations and the migration must be restarted manually. |
+| A retry always produces a single-VM plan | Retrying one VM from a plan that covered several VMs moves that VM into a plan of its own. The remaining VMs stay in the original plan and are unaffected. |
+| Bulk retry cannot change configuration | **Retry Selected** restarts each migration with its existing configuration. To change settings, retry the migration individually. |
+| Credentials and source cluster are locked | A retry cannot change the VMware or OpenStack credentials or the source cluster. Create a new migration instead. |
```
OmkarDeshpande7, line 254: @sarika-pf9 why do we have this limitation?

sarika-pf9: yes. https://github.com/platform9/vjailbreak/pull/1365

OmkarDeshpande7: @sarika-pf9 "Credentials and source cluster are locked". why do we have this restriction?

sarika-pf9: The VM being retried is identified by its VMwareMachine object, whose name and labels embed the source credential name. Retry reuses that key for the new plan and for the flavor patch, so a different credential means the object no longer resolves — the retry would fail or bind to the wrong VM in another vCenter.


The VM is already fixed. Choosing another cluster doesn't relocate the VM, it just desyncs the mapping step from the VM's real networks/datastores.

OmkarDeshpande7: "source credential name" - not quite right, its datastore/cluster something, but agreed its locked.

we should be able to change the target cluster though, that is a very valid use case where user accidentally used wrong cluster/tenant and hence the migration failed.
The retry option being available implies that the migration has failed. 

We should be able to allow target creds change, we can handle network/datastore mappings de-sync, that is not an issue. 

Please raise a ticket if your feel my request is valid. we will take it up as a follow up change

sarika-pf9: Target PCD cluster is editable on retry.
rest of are handled in #2273 #2274

